PR #53, opened 2014-03-13
Configurable default formatter

abinici:
I've added the option to configure a default formatter through the configuration-file which will then preselect this formatter in the dropdown-list.

I assume most users will prefer some specific formatter and having pre-selected this formatter would therefore improve the user-experience of this bundle.


sagikazarmark:
As I can see automatic hiding is already implemented. Default formatter would still be useful.

OskarStark:
Yes, the hiding is already implemented.

:+1:  for default formatter, otherwise the first formatter in the list is the default :(

sagikazarmark:
To preserve BC default value must be optionally checked. If not set, the first in the list need to be used. In a major version default option can be made mandatory.

sagikazarmark:
Oh, just noticed @OskarStark wrote essentially the same. :wink:

OskarStark:
:smile:

sagikazarmark:
Based on other bundles, the options should be called `default_formatter`.

Should the value be handled in the dependency injection part? I mean: process the configuration, and set the first value the default so it does not have to be checked in the code. When moving up a major version, that check can simply be removed in one place.

OskarStark:
:+1: for ```default_formatter```

Yes, please handle the this config option in the dependency part to avoid to check it everywhere in the code

rande:
This PR need to be rebased before being reviewed.

abinici:
I committed a sin. I wanted to change my author information in my commits, specifically I wanted to change my email address, by using `git filter-branch` and then did a force push. Now it looks like this PR contains a large number of commits. Can this be fixed? If there is no easy way to fix this maybe I should just close this PR and create a new one?

abinici:
If you guys don't mind I would like to close this PR and create a fresh one, because the current PR has become messy.

Based on the comments / proposals I will implement this feature with these requirements:

- The config for setting a default formatter will be called `default_formatter`
- For backward compatibility reasons the config setting will only be optional. When not set the first formatter in the list be used.
- This config option will be handled in the dependency injection part.

@sagikazarmark How is the automatic hiding implemented? Has it been implemented in a later commit in the main repo?

sagikazarmark:
@abinici I am already working on this PR. Writting tests right now.

Hiding is implemented in the FormatterType.

sagikazarmark:
Pushed the new PR.

abinici:
@sagikazarmark Sounds good :)

Can I close this PR then?

sagikazarmark:
Sure.

PR #76, opened 2014-09-14
Update FormatterType.php

pirasterize:
To avoid to override the FormatterType.php just for this property (ckeditor_context). Even without a default configuration in sonata_formatter.yml seems work perfectly.


OskarStark:
ping @rande

rande:
There is no need to overwrite the type, just define the option when you define the widget in your form type.

PR #82, opened 2014-11-13
fix for closure serializer exception "Serialization of 'Closure' is not ...

rmzamora:
fix for closure serialization exception "Serialization of 'Closure' is not allowed"


rande:
This issue has been fixed, thanks for this correct solution.

PR #92, opened 2015-03-08
Implement possibility to add plugins in CKEditor

roukmoute:
Proposal for an implementation of external plugins.

Example:
```php
$formMapper->add('content', 'sonata_formatter_type', [
    'ckeditor_plugins' => array(
        'pbckcode' => array(
            'path' => '/bundles/app/pbckcode/',
            'file' => 'plugin.js' // Optional if the default filename is plugin.js
        )
    )
])
```

### DIFF
--- a/Form/Type/FormatterType.php
+++ b/Form/Type/FormatterType.php
@@ -143,6 +143,9 @@ class FormatterType extends AbstractType
             $ckeditorConfiguration = array_merge($ckeditorConfiguration, $contextConfig);
         }
 
+        $view->vars['ckeditor_plugins']  = $options['ckeditor_plugins'];
+        $ckeditorConfiguration['extraPlugins'] = implode(',', array_keys($options['ckeditor_plugins']));
+
         $view->vars['ckeditor_configuration'] = $ckeditorConfiguration;
         $view->vars['ckeditor_basepath'] = $options['ckeditor_basepath'];
 
@@ -175,6 +178,7 @@ class FormatterType extends AbstractType
                  '-', 'Image', 'Link', 'Unlink', 'Table'),
                  array('Maximize', 'Source')
             ),
+            'ckeditor_plugins'          => array(),
             'ckeditor_basepath'         => 'bundles/sonataformatter/vendor/ckeditor',
             'ckeditor_context'          => null,
             'format_field_options'      => array(
--- a/Resources/views/Form/formatter.html.twig
+++ b/Resources/views/Form/formatter.html.twig
@@ -49,6 +49,14 @@
                         appendClass = 'html';
                         break;
                     case 'richhtml':
+                        {%- for name, plugin in ckeditor_plugins if ckeditor_plugins is defined -%}
+                            CKEDITOR.plugins.addExternal('{{ name }}', '{{ plugin.path }}',
+                                    {%- if plugin.file is defined -%}
+                                        '{{ plugin.file }}'
+                                    {%- else -%}
+                                        'plugin.js'
+                                    {%- endif -%});
+                        {% endfor %}
                         {{ source_id }}_rich_instance = {{ ckeditor_replace(form.children[source_field].vars.id, ckeditor_configuration) }};
 
                 }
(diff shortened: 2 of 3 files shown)


roukmoute:
@pulzarraider && @rande

Form/Type/FormatterType.php
```diff
@@ -175,6 +180,7 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
                  '-', 'Image', 'Link', 'Unlink', 'Table'),
                  array('Maximize', 'Source')
             ),
+            'ckeditor_plugins'           => null,
```
rande: Can you use ``array`` as default value ?

Form/Type/FormatterType.php
```diff
@@ -143,6 +143,11 @@ public function buildView(FormView $view, FormInterface $form, array $options)
             $ckeditorConfiguration = array_merge($ckeditorConfiguration, $contextConfig);
         }
 
+        if (is_array($options['ckeditor_plugins'])) {
```
rande: no need to check the type, just set the valid value

Resources/views/Form/formatter.html.twig
```diff
@@ -49,6 +49,16 @@
                         appendClass = 'html';
                         break;
                     case 'richhtml':
+                        {%- if ckeditor_plugins is defined -%}
```
rande: remove the ``if``

roukmoute:
thanks @rande

Resources/views/Form/formatter.html.twig
```diff
@@ -49,6 +49,14 @@
                         appendClass = 'html';
                         break;
                     case 'richhtml':
+                        {%- for name, plugin in ckeditor_plugins if ckeditor_plugins is defined -%}
```
rande: Cannot we add the plugin to the current instance, and not the global configuration ? if the configuration is global then 2 ckeditor on the same page will share the some set of plugins,.

greg0ire: @rande @roukmoute tells me there is no global configuration change with this P.R. The code you see here is supposed to act on the current instance only. What makes you think otherwise ?

Resources/views/Form/formatter.html.twig
```diff
@@ -49,6 +49,14 @@
...
+                            CKEDITOR.plugins.addExternal('{{ name }}', '{{ plugin.path }}',
```
rande: @greg0ire This line ;)

greg0ire: @rande :  oh ok. After discussing with @roukmoute , it appears that modifying the current instance is totally possible. @roukmoute is going to try that soo.

Resources/views/Form/formatter.html.twig
```diff
@@ -49,6 +49,14 @@
...
+                                    {%- if plugin.file is defined -%}
+                                        '{{ plugin.file }}'
+                                    {%- else -%}
+                                        'plugin.js'
+                                    {%- endif -%});
+                        {% endfor %}
                         {{ source_id }}_rich_instance = {{ ckeditor_replace(form.children[source_field].vars.id, ckeditor_configuration) }};
```
greg0ire: @rande : @roukmoute tells me he will try to do that here.

rande: I think this might help, http://docs.ckeditor.com/#!/api/CKEDITOR.config-cfg-extraPlugins

numerogeek:
hi ! 
Why it's not merged? That could be useful.

rande:
@numerogeek just read comments, you will have the response.

roukmoute:
sorry it's my fault I have not taken the time to continue this merge.

alexkvak:
@roukmoute could you finish this merge?

roukmoute:
Hi,

Sorry I do not use SonataAdminBundle since few months.  
If you prefer to finish the job you could.

if you really want I'll do it, I will take the time to do it.

alexkvak:
@roukmoute I found another solution #106

PR #100, opened 2015-06-13
Fixed JS error when switching formatter from ckeditor

andythorne:
When switching from ckeditor, a JS error is thrown as  `{{ source_id }}_rich_instance` is not destroyed, only unbound.

This fix checks the status of `{{ source_id }}_rich_instance` to check the instance has been destroyed

### DIFF
--- a/Resources/views/Form/formatter.html.twig
+++ b/Resources/views/Form/formatter.html.twig
@@ -20,8 +20,12 @@
                 return;
             }
 
+            var isInstanceActive = isInstanceActive || function(instance){
+                return instance && instance.status !== "destroyed";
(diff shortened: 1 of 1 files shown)


andythorne:
It builds fine - composer is failing to download dependencies.

Resources/views/Form/formatter.html.twig
```diff
@@ -20,8 +20,12 @@
                 return;
             }
 
+            var isInstanceActive = isInstanceActive || function(instance){
```
OskarStark: i like ```hasActiveInstance```, can you rename the variable?

PR #108, opened 2015-09-27
Add default formatter configuration

sagikazarmark:
Reimplementation of #53


sagikazarmark:
I am not sure why the documentation file I modified has that big diff.

DependencyInjection/SonataFormatterExtension.php
```diff
@@ -56,7 +56,19 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('ckeditor.xml');
         }
 
+        // TODO: To be removed when major version is changed
```
OskarStark: After we merged this PR, please create an issue and set the next major version as milestone, that we don't forget this!

Thank you @sagikazarmark

sagikazarmark: Should I create the issue?

DependencyInjection/Configuration.php
```diff
@@ -23,6 +23,7 @@ public function getConfigTreeBuilder()
 
         $rootNode
             ->children()
+                ->scalarNode('default_formatter')->end() // TODO: make it required when the major version is changed
```
OskarStark: an we validate the given input here, as we know which formatters are available, right?

```ifInArray()```

File: Formatter/Pool.php [hunk 3]
```diff
@@ -20,11 +20,16 @@ class Pool
 {
     protected $formatters = array();
 
+    protected $defaultFormatter;
+
     protected $logger;
 
-    public function __construct(LoggerInterface $logger = null)
+    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
```
OskarStark: Do we have an interface for type hint?

DependencyInjection/SonataFormatterExtension.php
OskarStark: only if this PR gets merged :smile:

Formatter/Pool.php
sagikazarmark: Interface for what? It is a string value.

DependencyInjection/Configuration.php
sagikazarmark: Validation is done in the extension before creating the formatter pool.

Formatter/Pool.php
OskarStark: ok, i thought there was a wrapper class or something like this. Thank you!

sagikazarmark: It works the same as in case of default_context in MediaBundle.

DependencyInjection/Configuration.php
OskarStark: Sorry my question is why there and not in the config before.


```php
        $supportedFormaters = array('foo', 'bar');
```
```php
            ->scalarNode('default_formatter')
                ->info(sprintf(
                    'These following formatters are supported: %s',
                    implode(', ', $supportedFormatters)
                ))
                ->validate()
                    ->ifNotInArray($supportedFormaters)
                    ->thenInvalid('The formatter "%s" is not supported. Please choose one of ('.implode(', ', $supportedDrivers).')')
                ->end()
            ->end()
```

the benefit ist that when we use ```config:dump-reference``` we have the info right there.

Formatter/Pool.php
OskarStark: :+1:

DependencyInjection/Configuration.php
sagikazarmark: Because the list of supported formatters are also defined in the config. Supported formatters are the keys defined in the formatters section.

sagikazarmark: Everything is done the same as in MediaBundle except from a few BC preserving code.

OskarStark: ok

OskarStark:
@sagikazarmark can you please check again why the doc file is modified that much?

sagikazarmark:
If I have to guess, it's line ending. Has anyone modified this file from windows?

If you take a look at the rich diff, you can see that only the added line is modified, which confirms my suspicion that line ending is change. (In this case the line ending is the unix default now, which should be used anyway, screw windows :stuck_out_tongue: )

OskarStark:
can you please provide the line endings cs fix as a separate PR?

sagikazarmark:
In that case I would have to not provide documentation for this PR. I can repair line endings for the rest of files in a separate PR if you wish, but should I provide documentation for this PR in that PR?

OskarStark:
i would recommend to fix the line endings in a separate PR, then we merge this cs fix PR and you can rebase your feature PR onto this.

what do you think?

sagikazarmark:
OK

OskarStark:
:+1:

sagikazarmark:
Hm, PR sent, but not sure what happens, because line endings are not corrected.

sagikazarmark:
Maybe it was indentation. Anyway, I can't find any problem with the file now. Can't you just merge it if everything seems to be in order? :wink:

OskarStark:
you provided a PR and there was no diff?

sagikazarmark:
Nope, you just merged it. ;)

OskarStark:
strange, but i don't want to merge it this way

can you add me as a collaborator to your repository?

sagikazarmark:
Added. Feel free to do anything, I can't really find the issue.

OskarStark:
hmm, i tried to fix it, but its still the same diff...

i think we have to merge it this way...

sagikazarmark:
:+1: I don't like it either, but I have no better idea.

OskarStark:
Thank you @sagikazarmark

sagikazarmark:
Hm. We should decide what should happen when no formatters are configured:

- We could say that default formatter can be null if no formatters are configured.
- It is an exceptional case, formatters should always be configured. Isn't it BC incompatible to throw an exception from now? If so, error should be triggered instead and option one should be chosen temporarily.

Also, we could check how the media bundle solves this question, I copied the whole default solution from there.

The only difference between the two solution is that for BC reasons, default formatter is not required.....for now. But it will be. In that case, this issue will be invalid. So we can say that it only exists because of BC reasons.

ping @Soullivaneuh

OskarStark:
can you move your last comment to a new Issue and reference this PR please?

Thank you

PR #113, opened 2015-10-06
[MOD] Apply configuration of SonataAdmin make cs

Th3Mouk:

PR #114, opened 2015-10-06
[ADD] Configuration of SonataAdmin make cs

Th3Mouk:


Soullivaneuh:
Was not `.php_cs` on `2.3` branch? That's strange.

Th3Mouk:
The file was added on master in june so not really weird :blush:. That's why i've do the #113 on master

PR #115, opened 2015-10-06
[UPT] Gitignore

Th3Mouk:
This folders mustn't be added on the .gitignore of the project, they're relative to the user environnement. Best practices advise to add the configuration of the user environnement in its own `.gitignore_global`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,6 @@
-.idea
-.DS_Store
 build
 phpunit.xml
 Resources/doc/_build/*
-nbproject
 coverage
 composer.lock
 vendor


Soullivaneuh:
Agree.

What about .DS_Store? This is only for mac env AFAIK.

Th3Mouk:
I've check but i'm on mac and i don't have this file, i'm agree but i wasn't totally certain

Soullivaneuh:
I'm globally OK for all, but for `.DS_Store` I would like to have opinion of other @sonata-project/contributors.

dbu:
my take on |.DS_Store| and other development environment artifacts
(.idea, whatever netbeans and all the editors do, ...) should be in the
global git ignore, not in project specific. it pollutes the .gitignore

Soullivaneuh:
That what I was thinking, thanks @dbu.

Th3Mouk:
@Soullivaneuh able to merge ?

PR #116, opened 2015-10-06
[FIX] Deprecated Twig functions usage

Th3Mouk:


Twig/Extension/TextFormatterExtension.php
```diff
@@ -50,7 +52,7 @@ public function getTokenParsers()
     public function getFilters()
     {
         return array(
-            'format_text' => new \Twig_Filter_Method($this, 'transform'),
+            'format_text' => new \Twig_SimpleFilter($this, 'transform'),
```
Soullivaneuh: Did you check Twig minimum requirement?

Twig/Extension/TextFormatterExtension.php
```diff
@@ -20,6 +20,8 @@ class TextFormatterExtension extends \Twig_Extension
 {
     protected $pool;
 
+    protected $environment;
```
Soullivaneuh: Please add a php doc block for this.

Yes I know. No php doc on another property. That is not a reason! :-)

Th3Mouk: :laughing:

Twig/Extension/TextFormatterExtension.php
Th3Mouk: yes it's 1.12.0, the date release is 2013-01-08 13:42 UTC, now we are 1.22

Soullivaneuh: So let's take profit to improve the requirement on composer: https://github.com/sonata-project/SonataFormatterBundle/blob/ae7b4084a14f24348537bd58a62e0a06d0d2efb2/composer.json#L23

;-)

Th3Mouk: Done

Soullivaneuh:
Could you rename commit comment and pr title?

> Fix deprecated Twig functions usage

Th3Mouk:
@Soullivaneuh :+1:

Soullivaneuh:
Thank you @Th3Mouk

wodka:
this is not good -> Twig_SimpleFilter takes as a first argument the name of the filter.

ErrorHandler ->handleError ('2', 'Illegal offset type', '/.../app/cache/dev/classes.php', '3669', array('extension' => object(TextFormatterExtension), 'name' => object(TextFormatterExtension), 'filter' => object(Twig_SimpleFilter)))

Twig/Extension/TextFormatterExtension.php
wodka: for this to work it would have to look like: \Twig_SimpleFilter('format_text', array($this, 'transform'))

Soullivaneuh:
@wodka, indeed. How I missed that? Thanks, will correct quickly.

rande:
@Soullivaneuh BTW, the  SonataCi / Sandbox / PHPUnit  was failing ;)

https://gist.github.com/rande/d68f7a6a266eb8f2fc8a

Soullivaneuh:
> @Soullivaneuh BTW, the SonataCi / Sandbox / PHPUnit was failing ;)

Well, one time it was always failling. I was used to ignore it.

Is that stable now?

rande:
@Soullivaneuh it is stable ;)

PR #124, opened 2015-11-10
Block translation

core23:

PR #128, opened 2015-11-14
Fix wrong block name

core23:


rande:
This is quite annoying bug ... and this is a major BC break.

core23:
Yes, this would be a BC break and could cause problems, but the risk is extreme low. There is also a problem with the ``EmptyBlockService`` which shares the same name.

As far as I know, the only usage for the name is for translation which was fixed by https://github.com/sonata-project/SonataFormatterBundle/pull/124. So I would prefer an upgrade notice and make an exception for this bug.

rande:
It is also used by the PageBundle to reference block type... but broken now ...

core23:
Can you show me the location? Couldn't find anything.

rande:
there is no link, there is a field in the page table call type. And type reference the block code ...

core23:
But the type comes from the service id, not the block name. If it would come from the name, you could not use an ``EmptyBlockService`` or `FormatterBlockService`` because they share the same name.

rande:
Good point1

PR #130, opened 2015-11-24
remove deprecated initRuntime method from twig extension, and remove unused instance variable for environment

shieldo:
Changes for compatibility with Twig 2.

There's only a danger here if someone has subclassed this extension class instead of using it directly.

### DIFF
--- a/Twig/Extension/TextFormatterExtension.php
+++ b/Twig/Extension/TextFormatterExtension.php
@@ -24,24 +24,11 @@ class TextFormatterExtension extends \Twig_Extension implements \Twig_Extension_
     protected $pool;
 
     /**
-     * @var \Twig_Environment
-     */
-    protected $environment;
-
-    /**
      * @param Pool $pool
      */
     public function __construct(Pool $pool)
     {
         $this->pool = $pool;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function initRuntime(\Twig_Environment $environment)
-    {
-        $this->environment = $environment;
     }
 
     /**


OskarStark:
@Soullivaneuh can you please verify this PR?

gsdevme:
I actually went to fix this myself without looking at existing PRs (https://github.com/sonata-project/SonataFormatterBundle/pull/131), can we get some movement.

Soullivaneuh:
@shieldo This PR should target 2.3 branch. Could you please manage it?

BTW, could you also improve `twig/twig` requirement? https://github.com/sonata-project/SonataFormatterBundle/blob/aee9121f844dc8b412831807e7da6e26dcead713/composer.json#L22

gsdevme:
Why does this have to target 2.3? It could be merged into any and all. its unused code.. just so happens that Twig 2.0 is upset by it.

core23:
@Soullivaneuh Shouldn't this get scheduled for the next major release? Someone could have used the ``environment`` property.

gsdevme:
I mean I assume its there as more of a copy/paste bug.. that's why I would think it needs removed from all, as a fix.

rande:
I added the interface check so the code should be compliant with twig 2. : https://github.com/sonata-project/SonataFormatterBundle/commit/dfcec12d58cb68dedb65c864ba9850f5dca9dc8e

gsdevme:
@rande As I noted here (https://github.com/sonata-project/SonataFormatterBundle/pull/131) thats not really the best way to let Twig know though. 

"As of Twig 1.23, the Twig_ExtensionInterface::initRuntime() method is deprecated. You have two options to avoid the deprecation notice: if you implement this method to store the environment for your custom filters, functions, or tests, use the needs_environment option instead; if you have more complex needs, explicitly implement Twig_Extension_InitRuntimeInterface (not recommended)."
http://twig.sensiolabs.org/doc/deprecated.html

rande:
@Soullivaneuh points is to avoid BC breaks for people using the current code in an extended class. No sure this situation really happen for this class. So I would be ok to merge as it. 

As a side note: @gsdevme "not recommended" does not mean "do not use it". There are many pieces of code in Sonata that are not recommended but still there to avoid to many BC break. Here one.

gsdevme:
@rande Well for the sake of the point if you think this is BC then it does need targeted against 2.3 I would of thought? 

We are talking about dead code being BC though? seems madness, granted some very odd use case where a user has extended it could run into problems.. Just wanted to make the "not recommended" clear and put it onto the table for discussion, not a holy grail rule for me either.

rande:
@gsdevme I am not against merging this PR. I am just listening to other core contributors voices. But in this case, I think you are right. 2.3 is the current branch anyway. @Soullivaneuh Can we merge this  PR ?

Soullivaneuh:
> @Soullivaneuh points is to avoid BC breaks for people using the current code in an extended class.

Not really for this case even if it's true.

I say this should be merge on 2.3 to have deprecation fix on a stable release.

Soullivaneuh:
BTW @rande I think this kind of classes should be `final` on a next major release to get rid of "possible extended" class BC handling.

shieldo:
@Soullivaneuh Thanks - I've created PR #135, so can close this one.  In terms of worries about BC - I think these are over-laboured considering how simple this extension is - I was really just mentioning it in terms of completeness.

OskarStark:
ok lets see what @Soullivaneuh @rande will finally say with the BC break

shieldo:
Sure. @Soullivaneuh @rande this has been merged against the 2.3 branch - I assume we can now merge it to master?

OskarStark:
@shieldo can you please rebase again, maybe something in the travis stuf is changed, as the master branch is green!

ping me after that and i will merge this PR

shieldo:
@OskarStark Hi! I've rebased, and everything looks green now. Thanks!

OskarStark:
Thank you @shieldo

PR #131, opened 2015-12-09
Remove unused Twig_Environment

gsdevme:
Using the `Twig_Extension_InitRuntimeInterface` is not recommended, since the `transform` method is not actually using the `Twig_Environment` Im not sure why its needed? I can add it back via the preferred method of through the options though if you explain the reason for keeping it. 

- [ ] change the unit tests


gsdevme:
@todo need to change the unit tests

PR #136, opened 2016-01-14
Update installation.rst

nottavi:
Without those lines in config.yml I can't get ckeditor to be shown (nor markdown ...) - I found those line in a forum post but not in the doc. So I suggest we add it if it is a right way to do (It made it on my side)


OskarStark:
Thank you @nottavi

PR #142, opened 2016-02-11
Added missing dependencies

core23:
The ``SonataAdminBundle`` and ``SonataMediaBundle`` should be ``required-dev``, because there is an integration in the admin backend.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,12 +23,14 @@
         "twig/twig": "~1.12",
         "knplabs/knp-markdown-bundle": "~1.4",
         "egeloen/ckeditor-bundle": "^2.5|^3.0|^4.0",
-        "sonata-project/block-bundle": "~2.2,>=2.2.1",
-        "sonata-project/core-bundle": "~2.3,>=2.3.1"
+        "sonata-project/block-bundle": "^2.2.1",
+        "sonata-project/core-bundle": "^2.3.1"
     },
     "require-dev": {
         "symfony/validator": "~2.3|~3.0",
-        "symfony/phpunit-bridge": "~2.7|~3.0"
+        "symfony/phpunit-bridge": "~2.7|~3.0",
+        "sonata-project/admin-bundle": "~2.3",
+        "sonata-project/media-bundle": "~2.2"
     },
     "autoload": {
         "psr-4": { "Sonata\\FormatterBundle\\": "" }
@@ -37,5 +39,7 @@
         "branch-alias": {
             "dev-master": "2.4.x-dev"
         }
-    }
+    },
+    "minimum-stability": "dev",
+    "prefer-stable": true
 }
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,5 @@
         "branch-alias": {
             "dev-master": "2.4.x-dev"
         }
-    },
-    "minimum-stability": "dev",
-    "prefer-stable": true
+    }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
     - php: 7.0
       env: TARGET=docs
     - php: 5.3
-      env: COMPOSER_FLAGS="--prefer-lowest --prefer-stable"
+      env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.6
       env: SYMFONY_VERSION=2.3.*
     - php: 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
     - php: 7.0
       env: TARGET=docs
     - php: 5.3
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest --profile -vvv"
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -23,12 +23,15 @@
         "twig/twig": "~1.12",
         "knplabs/knp-markdown-bundle": "~1.4",
         "egeloen/ckeditor-bundle": "~2.2",
-        "sonata-project/block-bundle": "~2.2,>=2.2.1",
-        "sonata-project/core-bundle": "~2.3,>=2.3.1"
+        "sonata-project/admin-bundle": "~2.3",
+        "sonata-project/block-bundle": "~2.2",
```
OskarStark: the `>=2.2.1` is not imortant?

composer.json
```diff
@@ -23,12 +23,15 @@
...
+        "sonata-project/core-bundle": "~2.3",
```
OskarStark: sanem here?

core23: Copy & paste issue ;)

composer.json
```diff
@@ -23,12 +23,15 @@
         "twig/twig": "~1.12",
         "knplabs/knp-markdown-bundle": "~1.4",
         "egeloen/ckeditor-bundle": "~2.2",
+        "sonata-project/admin-bundle": "~2.3",
         "sonata-project/block-bundle": "~2.2,>=2.2.1",
-        "sonata-project/core-bundle": "~2.3,>=2.3.1"
+        "sonata-project/core-bundle": "~2.3,>=2.2.1",
```
OskarStark: 2.3.1 not 2.2.1

OskarStark: :smile:

core23:
Next try ;)

composer.json
```diff
@@ -23,12 +23,15 @@
         "twig/twig": "~1.12",
         "knplabs/knp-markdown-bundle": "~1.4",
         "egeloen/ckeditor-bundle": "~2.2",
+        "sonata-project/admin-bundle": "~2.3",
         "sonata-project/block-bundle": "~2.2,>=2.2.1",
-        "sonata-project/core-bundle": "~2.3,>=2.3.1"
+        "sonata-project/core-bundle": "~2.3,>=2.3.1",
```
OskarStark: if you put this to the end, it would not occur as a change, because of the colon

core23: But I love alphabetical order ;)

OskarStark: fine for me :+1:

core23:
Travis seems broken for this PR... :/

core23:
@greg0ire Do you have any idea why composer hangs up at the ``prefer-lowest`` task.

greg0ire:
Woops didn't see your question. Try removing `minimal-stability`

core23:
Same result :( @greg0ire

greg0ire:
Maybe `--profile -vvv` would help ?

Soullivaneuh:
@core23 Could you try to ass `composer selfupdate --stable`?

Ref: https://github.com/travis-ci/travis-build/pull/718

greg0ire:
I'm reproducing the problem locally with a stable version (1.0.3)

greg0ire:
I'll try to bisect it to some constraint.

greg0ire:
And the culprit is… `sonata-project/media-bundle` !

greg0ire:
The media bundle requires `symfony/symfony`. This looks a bit strange.

greg0ire:
@core23 : why exactly would the media bundle be necessary in dev ?

Soullivaneuh:
> The media bundle requires symfony/symfony. This looks a bit strange.

I think this is fixed on master, but not on old version.

core23:
> why exactly would the media bundle be necessary in dev ?

There are some deps in:
- https://github.com/sonata-project/SonataFormatterBundle/blob/master/Controller/CkeditorAdminController.php#L41-L68 
- https://github.com/sonata-project/SonataFormatterBundle/blob/master/Resources/config/ckeditor.xml#L13

greg0ire:
Hum indeed… maybe some of the fixes in master should be ported to 2.3 in the media-bundle, and then it should be tagged, and you would require this tag.

greg0ire:
I get the very same out of memory error when trying to run `composer update --prefer-dist --no-interaction --prefer-lowest --profile -vvv` on the 2.3 branch. I'm going to try to backport things from master, and when it is merge, the build should pass here too…

greg0ire:
Also, there is a "circular" reference (the package require each other, but in dev only, hence the double quotes). Dunno if "circular" references are really a problem…

greg0ire:
Sadly, I get the same error on the master branch of the media bundle too…

greg0ire:
Configured my swap, and it seems it is not a memory problem after all… my CPUs are working on the problem as we speak and there is no sign of them getting to the solution… it works fine in the media bundle, but still not on the formatter bundle, even after backporting some fixes to a branch on my fork and using an inline alias to reference this branch.

Soullivaneuh:

Regards.

PR #145, opened 2016-02-18
Compatibility with SF 3 and BC with SF <= 2.8

idchlife:

PR #110, opened 2015-10-01
[UPT] CKEditor dependency to ~3.0

Th3Mouk:
Upgrade `egeloen/ckeditor-bundle` to new major ~3.0

It's BC break, cause the `ckeditor_replace` tag has been renamed, so projects where the template `formatter.html.twig` has been overridden must be updated.

IMO sadly this commit can't be add to 2.3.x

PR #146, opened 2016-02-24
FormatterType Support SF3

rafaelcalleja:
add to #145 sonata_formatter_type support in SF3

### DIFF
--- a/Tests/Form/Widget/BaseWidgetTest.php
+++ b/Tests/Form/Widget/BaseWidgetTest.php
@@ -1,0 +1,150 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\FormatterBundle\Tests\Form\Widget;
+
+use Ivory\CKEditorBundle\Twig\CKEditorExtension;
+use Symfony\Bridge\Twig\Extension\FormExtension;
+use Symfony\Bridge\Twig\Extension\TranslationExtension;
+use Symfony\Bridge\Twig\Form\TwigRenderer;
+use Symfony\Bridge\Twig\Form\TwigRendererEngine;
+use Symfony\Bridge\Twig\Tests\Extension\Fixtures\StubFilesystemLoader;
+use Symfony\Bundle\FrameworkBundle\Tests\Templating\Helper\Fixtures\StubTranslator;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\Form\Test\TypeTestCase;
+
+/**
+ * Class BaseWidgetTest.
+ */
+abstract class BaseWidgetTest extends TypeTestCase
+{
+    /**
+     * @var FormExtension
+     */
+    protected $extension;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        $rendererEngine = new TwigRendererEngine(array(
+            'formatter.html.twig',
+            'form_div_layout.html.twig',
+        ));
+
+        if (class_exists('Symfony\Component\Form\Extension\Core\Type\RangeType')) {
+            $csrfManagerClass = 'Symfony\Component\Security\Csrf\CsrfTokenManagerInterface';
+        } else {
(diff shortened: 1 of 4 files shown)


OskarStark:
LGTM

greg0ire:
And the commit adds tests… great work!

Tests/Form/Widget/BaseWidgetTest.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tearDown()
+    {
+        parent::tearDown();
+
+        $this->extension = null;
```
core23: IMO, we don't need this

Tests/Form/Widget/BaseWidgetTest.php
```diff
@@ -0,0 +1,160 @@
...
+
+        $renderer = new TwigRenderer($rendererEngine, $this->getMock($csrfManagerClass));
+
+        $this->extension = new FormExtension($renderer);
+
+        $twigPaths = array(__DIR__.'/../../../Resources/views/Form');
+
+        //this is ugly workaround for different build strategies and, possibly,
+        //different TwigBridge installation directories
+        if (is_dir(__DIR__.'/../../../vendor/symfony/twig-bridge/Resources/views/Form')) {
```
core23: Could we move this to a seperate method? So we can use this for other tests too

greg0ire: What do you mean? It already is in a public method…

Tests/Form/Widget/BaseWidgetTest.php
dnoise: done

dnoise:
I've upgraded this PR with master because the Test was failing but the upgrade doesn't fix the Test.

Thank to the failed test I guess have detected a issue about versions of egeloen/ckeditor-bundle from the last merge #110 .

Too In my project i can see:
`The service "sonata.formatter.twig.ck_editor" has a dependency on a non-existent service "ivory_ck_editor.templating.helper".`

The service ivory_ck_editor.templating.helper was renamed in version 2.4 but composer.json is allowing version  2.2 
`"egeloen/ckeditor-bundle": "^2.2|^3.0"`

You can check it from [this commit](https://github.com/egeloen/IvoryCKEditorBundle/commit/5e8fc1094cffb3072373d72899a240a5958378ff)

To fix it and the test too simply need change composer.json to

`"egeloen/ckeditor-bundle": "~2.4|^3.0",`

maybe this change would be in a new PR or Issue?

greg0ire:
> maybe this change would be in a new PR or Issue?

I think it would. Can you do it? @Th3Mouk can you comment on this?

dnoise:
The firt version of egeloen/ckeditor-bundle with support for symfony/form:~3.0 is [4.0](https://github.com/egeloen/IvoryCKEditorBundle/blob/4.0.0/composer.json)

I guess we can support sf3 with something like this
"egeloen/ckeditor-bundle": "~2.4|^3.0|~4.0",

Th3Mouk:
I've started to develop the upgrade in october and there's not 4.x releases, i agree that the composer.json must be upgrade from `^2.2` to `^2.5` because when i've tested i was in `2.5.2`.

Il will make the upgrade of composer.

The ^4.0 upgrade seems really simple, @dnoise would you do the PR ?

Th3Mouk:
I've check it's ok for 4.x support there no BC, i've made PR to https://github.com/sonata-project/SonataFormatterBundle/pull/148

Tests/Form/Widget/BaseWidgetTest.php
core23: Can't we use a global WidgetTest class @greg0ire ? We have an existing here https://github.com/sonata-project/SonataAdminBundle/blob/master/Tests/Form/Widget/BaseWidgetTest.php

greg0ire: Formatter depends on Core, so I think we could try to factor out something into core. Maybe let it mature into something generic enough first, b/c tests must stay easy to write.

OskarStark:
can we merge this @Th3Mouk

Th3Mouk:
@OskarStark I can't pronounce on the testing part, that must be extract on a base class apparently.

@rafaelcalleja could you rebase please ? Tests are red actually

Soullivaneuh:

Regards.

PR #152, opened 2016-03-24
Do not overwrite the format provided by the user with the default format

mlueck88:
Symfony provides the possibility setting a default value for a choice type form field. The formatter type always overwrites the user setting with the default formatter of the formatter pool. This forbids for example to persist the format together with settings of Sonata blocks and to re-edit them later in the same format, as the form field will just jump back to 'text' format.

Instead, the given value should only be overwritten if no valid format was given by the user.

### DIFF
--- a/Form/Type/FormatterType.php
+++ b/Form/Type/FormatterType.php
@@ -67,7 +67,10 @@ class FormatterType extends AbstractType
             $options['format_field_options']['property_path'] = $formatField;
         }
 
-        $options['format_field_options']['data'] = $this->pool->getDefaultFormatter();
+        if (!array_key_exists('data', $options['format_field_options']) ||
+             !array_key_exists($options['format_field_options']['data'], $this->pool->getFormatters())) {
+            $options['format_field_options']['data'] = $this->pool->getDefaultFormatter();
+        }
 
         if (is_array($options['source_field'])) {
             list($sourceField, $sourcePropertyPath) = $options['source_field'];


greg0ire:
Please keep your commit subject (the first line of the commit message) under 50 chars.

Form/Type/FormatterType.php
```diff
@@ -67,7 +67,10 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             $options['format_field_options']['property_path'] = $formatField;
         }
 
-        $options['format_field_options']['data'] = $this->pool->getDefaultFormatter();
+        if (!array_key_exists($options['format_field_options']['data'], $this->pool->getFormatters()))
+        {
```
greg0ire: This violates PSR. Run `make cs` to see what I mean

mlueck88:
Does the 50 char limit apply only to the first commit? Either way, is it allowed to have a PR with 2 or more commits? Should I create a new PR to fix the initial commit message?

greg0ire:
> Does the 50 char limit apply only to the first commit?

It applies to every commit, it is here so that the short log (`git log --oneline`) looks nice.

> Either way, is it allowed to have a PR with 2 or more commits?

Yes, as long as both commits are independant from each other

> Should I create a new PR to fix the initial commit message?

No, but you do not need to. You need to rebase interactively (`git rebase --interactive` on the lastest master), mark the commits you want to reword with an r, and then force push the result to overwrite the commits you already pushed `git push --force`. Do not hesitate to ask for details if you are stuck.

greg0ire:
[This article](http://rakeroutes.com/blog/deliberate-git/) explains all this, BTW.

mlueck88:
Thanks. I forced push of the new commit.

greg0ire:
Congrats, you did it!

greg0ire:
Your PR message is great, BTW! Could you add it to your commit message, like this : 

```
Don't overwrite formatter selection by user

Symfony provides the possibility setting a default value for a choice type form
field. The formatter type always overwrites the user setting with the default
formatter of the formatter pool. This forbids for example to persist the format
together with settings of Sonata blocks and to re-edit them later in the same
format, as the form field will just jump back to 'text' format.
Instead, the given value should only be overwritten if no valid format was
given by the user.
```

mlueck88:
Ok, I added the details to the commit. Tests should run fine.

core23:
LGTM

Soullivaneuh:

Regards.

PR #154, opened 2016-04-12
[DEL] Quick clean up of unused parameter

Th3Mouk:


OskarStark:
is this a BC Break, or not, because of doing stuff in the extension? @greg0ire

greg0ire:
Yes it is a BC break, because the signature of a public function loses a mandatory argument.

greg0ire:
BTW, why is it even public ? @Th3Mouk : I think you should make it private.

Th3Mouk:
Ok so this PR must wait the creation of 4.x branch

greg0ire:
> Ok so this PR must wait the creation of 4.x branch

Indeed. But it should come soon…

OskarStark:
4.0? at least the next major.

and please add an upgrade notice

greg0ire:
> 4.0? at least the next major.

Not 4.0, 4.x, these are very different things ;) and I think 4.x will be the alias of master. 3.x will be created, not 4.x . When it is out, we will be able to merge this in master.

> and please add an upgrade notice

:+1: In that case, it is just an internal method that changes, but that will teach us to make method that should be private public ;) 
The upgrade notice title should read : "From 3.x to 4.x" or something like that.

Soullivaneuh:

Regards.

PR #166, opened 2016-05-22
Added missing dependencies

core23:
### Changelog

```markdown
### Added
- Added `SonataAdminBundle` to suggest 
- Added `SonataMediaBundle` to suggest 
```

### Subject

The ``SonataAdminBundle`` and ``SonataMediaBundle`` should be suggested, because there is an integration in the admin backend.

### To do

- [X] Add missing dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,14 @@
         "symfony/validator": "^2.3 || ^3.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0"
     },
+    "suggest": {
+        "sonata-project/admin-bundle": "For using the admin media browser.",
+        "sonata-project/media-bundle": "For using the admin media browser."
+    },
+    "conflict": {
+        "sonata-project/admin-bundle": "<3.0 || >= 4.0",
+        "sonata-project/media-bundle": "<3.0 || >= 4.0"
+    },
     "autoload": {
         "psr-4": { "Sonata\\FormatterBundle\\": "" }
     },


OskarStark:
LGTM, but travis failed

jordisala1991:
Versions should be ^3.0 I think

core23:
There are no stable versions yet

Soullivaneuh:
> There are no stable versions yet

@core23 admin-bundle has a stable version from time: https://packagist.org/packages/sonata-project/admin-bundle

And in any way, we should use 3.x.

For media-bundle, does the project have test requiring this dep? If not, it should be removed from `require-dev` for the moment.

core23:
I would add missing tests afterwards

Soullivaneuh:
> I would add missing tests afterwards

Same thing for `sonata-project/admin-bundle`?

In this case, could you add this dependency on another PR with test? In this case, we can merge this one first with the `suggest` section.

core23:
Moved the stuff to `suggest` and `conflict`.

composer.json
```diff
@@ -31,6 +31,14 @@
         "symfony/validator": "^2.3 || ^3.0",
         "symfony/phpunit-bridge": "^2.7 || ^3.0"
     },
+    "suggest": {
+        "sonata-project/admin-bundle": "For using the admin media browser.",
+        "sonata-project/media-bundle": "For using the admin media browser."
+    },
+    "conflict": {
+        "sonata-project/admin-bundle": "<3.0",
```
Soullivaneuh: ` || >=4.0`. Same for line below. :+1:

core23: Good point 👍

core23:
> For media-bundle, does the project have test requiring this dep? If not, it should be removed from require-dev for the moment.

Sure we should use `require-dev` only for running tests @Soullivaneuh ? Composer [docs](https://getcomposer.org/doc/04-schema.md#require-dev) say, that you could use them for development too. 

For example when running code inspection, I got "class not found errors" because it can't resolve the admin classes.

PR #169, opened 2016-05-22
Remove files and informations relative to 3.x branch

Th3Mouk:
This informations and files doesn't concern 4.x releases


Soullivaneuh:
:-1:

We have to keep at lest one major version before notes.

PR #181, opened 2016-05-30
Fix overwriting formatter selection

mlueck88:
### Changelog

```markdown
### Fixed
- Fix overwriting formatter selection
```

### Subject

Symfony provides the possibility to give a default value for a choice type form field when setting up a form. The formatter type is such choice field.
It is currently however overwritten with the default formatter of the formatter pool without warning.
A common use case is to persist the format together with other settings of e.g. Sonata blocks and to re-edit them later in the same format, but currently the form field will just jump back to 'text' format and render the block content unusable.
Instead, the given value should only be overwritten if no valid format was given by the user.

### DIFF
--- a/Form/Type/FormatterType.php
+++ b/Form/Type/FormatterType.php
@@ -67,7 +67,10 @@ class FormatterType extends AbstractType
             $options['format_field_options']['property_path'] = $formatField;
         }
 
-        $options['format_field_options']['data'] = $this->pool->getDefaultFormatter();
+        if (!array_key_exists('data', $options['format_field_options']) ||
+             !array_key_exists($options['format_field_options']['data'], $this->pool->getFormatters())) {
+            $options['format_field_options']['data'] = $this->pool->getDefaultFormatter();
+        }
 
         if (is_array($options['source_field'])) {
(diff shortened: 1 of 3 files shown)


mlueck88:
@core23 For test cases, should I modify the methods of the existing FormatterTypeTest class? Or better add new methods to it? Or do something different?

core23:
A new method which tests your modifications 👍

Soullivaneuh:
I think concrete examples should be added to the documentation.

mlueck88:
I added test cases with mock form builders which check that the form is indeed built with the user provided formatter, or, if none is given, with the default formatter. For the documentation, I added an example.

OskarStark:
Good job @mlueck88 

LTM 👍

Soullivaneuh:
I let a contributor using this bundle merge this PR.

Avtonom:
@Soullivaneuh ping

Soullivaneuh:
@Avtonom ?

Avtonom:
@Soullivaneuh I'm sorry, just noticed that this is already included in Release

Issue #125, opened 2015-11-11
using the simple formatter widget in block does not save content using block in sonata page bundle


lordrhodos:
Hi there,

I have a project using the sonata admin and page bundle. I've created a custom block defining a field as "sonata_simple_formatter_type". Rendering of the CKEditor is working flawless, but it does not save the correct content when I try to create or update the block in the pages composer mode (e.g. http:/example.com/app_dev.php/admin/sonata/page/page/48/compose)

When I access the the block directly via the crud admin form (e.g. http:/example.com/app_dev.php/admin/sonata/page/block/440/edit?composer=1) there is no problem at all and I can update and create the content as needed.

I was able to debug the php side, so I know that the request in the latter case does not contain the updates made in the ckeditor for this form field.

lordrhodos:
I was able to fix this bug by updating the **sonata_simple_formatter_type_widget** block in the *formatter.html.twig* template:

```
{% block sonata_simple_formatter_type_widget %}

    <!-- widget: sonata_simple_formatter_type_widget -->

    {{ form_widget(form) }}

    <script>
        jQuery(document).ready(function() {
            // This code requires CKEDITOR and jQuery MarkItUp
            if (typeof CKEDITOR === 'undefined' || jQuery().markItUp == undefined) {
                return;
            }

            var elms = jQuery('#{{ form.vars.id }}'),
		            editor = false;

            {% if format == 'textile' %}
                elms.markItUp(markitup_sonataTextileSettings);
            {% elseif format == 'markdown' %}
                elms.markItUp(markitup_sonataMarkdownSettings);
            {% elseif format == 'bbcode' %}
                elms.markItUp(markitup_sonataBBCodeSettings);
            {% elseif format == 'rawhtml' %}
                elms.markItUp(markitup_sonataHtmlSettings);
            {% elseif format == 'richhtml' %}
                editor = {{ ckeditor_replace(form.vars.id, ckeditor_configuration) }};
            {% endif %}

            var parent = elms.parents('div.markItUp');

            if (parent) {
                parent.addClass('{{ format }}');
            }

	        elms.parents("form").on('submit', function (event) {
		        if (editor) {
			        editor.updateElement();
		        }
	        });

        });
    </script>
{% endblock sonata_simple_formatter_type_widget %}
```

OskarStark:
@lordrhodos which version are you using?

if this is still a problem, can you please provide a PR? That would be awesome :+1:

[...]

PR #193, opened 2016-06-24
fix simple widget not updating in sonata page block #125

hug963:
Fixes https://github.com/sonata-project/SonataFormatterBundle/issues/125

## Changelog

```markdown
### Fixed
- Field with `sonata_simple_formatter_type` widget in a block in sonata page's composer doesn't update
```

## Subject

When inserting a `sonata_simple_formatter_type` field in a custom block in sonata page, the content never gets updated when saving.

### DIFF
--- a/Resources/views/Form/ckeditor.html.twig
+++ b/Resources/views/Form/ckeditor.html.twig
@@ -1,0 +1,7 @@
+{% for plugin_name, plugin in ckeditor_plugins %}
+	{{ ckeditor_plugin(plugin_name, plugin) }}
+{% endfor %}
+
+{{ ckeditor_widget(ckeditor_field_id, ckeditor_configuration, {
+	input_sync: true,
+}) }}
--- a/Resources/views/Form/formatter.html.twig
+++ b/Resources/views/Form/formatter.html.twig
@@ -103,7 +103,9 @@
             {% elseif format == 'rawhtml' %}
                 elms.markItUp(markitup_sonataHtmlSettings);
             {% elseif format == 'richhtml' %}
-                {{ ckeditor_widget(form.vars.id, ckeditor_configuration) }};
+                {{ ckeditor_widget(form.vars.id, ckeditor_configuration, {
+                    input_sync: true,
+                }) }}
             {% endif %}
 
             var parent = elms.parents('div.markItUp');
--- a/Resources/views/Form/formatter.html.twig
+++ b/Resources/views/Form/formatter.html.twig
@@ -40,13 +40,7 @@
                         appendClass = 'html';
                         break;
                     case 'richhtml':
-                        {% for plugin_name, plugin in ckeditor_plugins %}
-                            {{ ckeditor_plugin(plugin_name, plugin) }}
-                        {% endfor %}
-
-                        {{ ckeditor_widget(form.children[source_field].vars.id, ckeditor_configuration, {
-                            input_sync: true,
-                        }) }}
+                        {% include 'SonataFormatterBundle:Form:ckeditor.html.twig' with {'ckeditor_field_id': form.children[source_field].vars.id} %}
                 }
 
                 var parent = elms.parents('div.markItUp');
@@ -89,9 +83,7 @@
             {% elseif format == 'rawhtml' %}
                 elms.markItUp(markitup_sonataHtmlSettings);
             {% elseif format == 'richhtml' %}
-                {{ ckeditor_widget(form.vars.id, ckeditor_configuration, {
-                    input_sync: true,
-                }) }}
+                {% include 'SonataFormatterBundle:Form:ckeditor.html.twig' with {'ckeditor_field_id': form.vars.id} %}
             {% endif %}
 
             var parent = elms.parents('div.markItUp');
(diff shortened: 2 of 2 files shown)


hug963:
I changed the fix, using the built-in "input_sync".

It works perfectly and is quite clean.

OskarStark:
@core23 can you please verify this PR?

Resources/views/Form/formatter.html.twig
```diff
@@ -99,7 +99,9 @@
             {% elseif format == 'rawhtml' %}
                 elms.markItUp(markitup_sonataHtmlSettings);
             {% elseif format == 'richhtml' %}
-                {{ ckeditor_widget(form.vars.id, ckeditor_configuration) }};
+                {{ ckeditor_widget(form.vars.id, ckeditor_configuration, {
+                    input_sync: true,
```
Soullivaneuh: I'm not sure but this seems to be BC break (default option changed).

hug963: I just does the same thing as https://github.com/hug963/SonataFormatterBundle/blob/e7a0c994669914729b2d25945c9134751d3f6829/Resources/views/Form/formatter.html.twig#L23-L31
in the main widget, but with the built-in ckeditor way.
For now it is a bug as some forms will not update.
This change will not break the forms that already work.

hug963:
Do you know when or if you'll merge this? 
Let me know if there is anything I can do.

greg0ire:
ping @Soullivaneuh : stuck build

hug963:
This is basically the same issue that was fixed here:
https://github.com/sonata-project/SonataFormatterBundle/commit/97c1a79a5685848ae716e29988867a1a1885df71

maybe I should remove those lines and apply the same technique to the main widget as well?

Resources/views/Form/formatter.html.twig
rande: The code should be BC

greg0ire:
I think if you rebase this, the build might get unstuck

hug963:
Hi @greg0ire . The Build is indeed passing after rebase.

Resources/views/Form/formatter.html.twig
greg0ire: I think this is BC and harmless, see http://symfony.com/doc/current/bundles/IvoryCKEditorBundle/usage/textarea-sync.html

greg0ire:
LGTM

OskarStark:
> Yes please. Let's get rid of as much javascript in twig as we can.

👍

hug963:
I've updated the main widget, to use the built-in "input sync" method as well as the "ckeditor_destroy" twig method to clear the form.

We end up recreating most of the ckeditorbundle template twice:
https://github.com/egeloen/IvoryCKEditorBundle/blob/master/Resources/views/Form/ckeditor_widget.html.twig

Also, having to update two separate widgets leads to a lot of errors, the recent commit https://github.com/sonata-project/SonataFormatterBundle/commit/4493c4ca688ae98bfb71adbfd034fdd077f94b5c adds plugins to both types, but forgot to render it in the simple widget.

Maybe we should have a separate template and include it?

```twig
{# 'SonataFormatterBundle:Fragment:ckeditor_form.html.twig' #}

{% for plugin_name, plugin in ckeditor_plugins %}
    {{ ckeditor_plugin(plugin_name, plugin) }}
{% endfor %}

{{ ckeditor_widget(widget_id, ckeditor_configuration, {
    input_sync: true,
}) }}
```

This way we'd be sure there is no difference between the main and simple widgets anymore.

greg0ire:
> Maybe we should have a separate template and include it?

:+1: for less duplication, are you willing to do it?

hug963:
Done.

Resources/views/Form/ckeditor.html.twig
```diff
@@ -0,0 +1,7 @@
+{% for plugin_name, plugin in ckeditor_plugins %}
+	{{ ckeditor_plugin(plugin_name, plugin) }}
+{% endfor %}
+
+{{ ckeditor_widget(ckeditor_field_id, ckeditor_configuration, {
+	input_sync: true,
+}) }}
```
greg0ire: Please configure your IDE to always add a newline at end of file, this will make the red pictogram disappear.

greg0ire:
Please fixup commit 5 into commit 4 and I think we will be good to go.

hug963:
I've squashed the last two commits.

I can squash everything if you want to.

greg0ire:
> I can squash everything if you want to.

No, but I think you should also squash the second into the first. You should be left with 3 commits that have different purposes and could be used (and reverted) independently.

hug963:
Done, I agree it is quite clean this way.

greg0ire:
LGTM

OskarStark:
LGTM, too @greg0ire feel free to merge instead of squash here please

greg0ire:
Thanks @hug963 !

pix-art:
When will this be added to a new release?

greg0ire:
I'll ask @Soullivaneuh for a release rightaway!

pix-art:
Awesome! :-)

Soullivaneuh:
Here you are: https://github.com/sonata-project/SonataFormatterBundle/releases/tag/3.1.0

PR #211, opened 2016-08-17
Check symfony version when using choice_translation_domain

### LINKED ISSUE #210: Widget broken with PR #205
## Subject
This PR https://github.com/sonata-project/SonataFormatterBundle/pull/205 broke the main widget in the page composer.

The simple widget is not affected as it was not updated.

I get this error in the admin:

> The option "choice_translation_domain" does not exist. Defined options are: "action", "allow_extra_fields", "attr", "auto_initialize", "block_name", "by_reference", "cascade_validation", "compound", "constraints", "csrf_field_name", "csrf_message", "csrf_protection", "csrf_provider", "csrf_token_id", "csrf_token_manager", "data", "data_class", "disabled", "empty_data", "error_bubbling", "error_mapping", "extra_fields_message", "horizontal_input_wrapper_class", "horizontal_label_class", "horizontal_label_offset_class", "inherit_data", "intention", "invalid_message", "invalid_message_parameters", "label", "label_attr", "label_format", "label_render", "mapped", "max_length", "method", "pattern", "post_max_size_message", "property_path", "read_only", "required", "sonata_admin", "sonata_field_description", "sonata_help", "static_control", "translation_domain", "trim", "validation_groups", "virtual".

core23:
I am targetting this branch, because this is a patch.

Fixes #210

## Changelog

```markdown
### Fixed
 - Fixed usage of `choice_translation_domain` in `FormatterType`
```

## Subject

Checks symfony version when using this option.

### DIFF
--- a/Form/Type/FormatterType.php
+++ b/Form/Type/FormatterType.php
@@ -197,6 +197,15 @@ class FormatterType extends AbstractType
             $formatters[$code] = $translator->trans($code, array(), 'SonataFormatterBundle');
         }
 
+        $formatFieldOptions = array(
+            'choices' => $formatters,
+        );
+
+        // NEXT_MAJOR: Remove the method_exists hack when dropping support for symfony < 2.7
+        if (count($formatters) > 1 && method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
+            $formatFieldOptions['choice_translation_domain'] = false;
+        }
+
         $resolver->setDefaults(array(
             'inherit_data' => true,
             'event_dispatcher' => null,
@@ -213,10 +222,7 @@ class FormatterType extends AbstractType
             'ckeditor_basepath' => 'bundles/sonataformatter/vendor/ckeditor',
             'ckeditor_context' => null,
             'ckeditor_plugins' => array(),
-            'format_field_options' => array(
-                'choices' => $formatters,
-                'choice_translation_domain' => false,
-            ),
+            'format_field_options' => $formatFieldOptions,
             'source_field' => null,
             'source_field_options' => array(
                 'attr' => array('class' => 'span10 col-sm-10 col-md-10', 'rows' => 20),


greg0ire:
LGTM, thanks for the reactivity!

hug963:
I'm getting the error on Symfony 2.8.9 though.

greg0ire:
Strange… can you test the fix before we merge it?

hug963:
I found the problem, I only have richhtml defined as a formatter service, so the returned form is a hidden input, not a select, and it doesn't have the choice_translation_domain option...

greg0ire:
I seem to lack a lot of context to be able to understand.

> I only have richhtml defined as a formatter service,

As opposed to what?

> so the returned form is a hidden input, not a select

What are you talking about?

> and it doesn't have the choice_translation_domain option...

So you're experiencing a different bug, requiring a different fix?

hug963:
Can be fixed with checking if there is more than one formatter
```php
if (count($formatters) > 1 && method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
    $formatFieldOptions['choice_translation_domain'] = false;
}
```

this works for me.

greg0ire:
Oh ok I get it now.

hug963:
Cool, I'm sorry it was quite confusing.

core23:
Buy why can we pass `choices` as an option?

greg0ire:
Because it is unset : https://github.com/sonata-project/SonataFormatterBundle/blob/28c79f051a514a526938ebc9db1762364fed7bdd/Form/Type/FormatterType.php#L99 (which is quite confusing)

core23:
This looks like a pretty ugly hack. We should improve it.

greg0ire:
> This looks like a pretty ugly hack. We should improve it.

Yes, but in another PR, b/c it probably won't be BC. I think a good solution would be to not try to sound like the choice type  when in fact it is a `hidden` field . Meaning rename the option from `choices` to `formatters` or sth like that (hence the BC break)

core23:
> Yes, but in another PR, b/c it probably won't be BC. I think a good solution would be to not try to sound like the choice type  when in fact it is a `hidden` field . Meaning rename the option from `choices` to `formatters` or sth like that (hence the BC break)

Refs https://github.com/sonata-project/SonataFormatterBundle/issues/212

core23:
Ping @greg0ire

greg0ire:
Thanks @core23 ! This PR is small enough that we can tolerate having one LGTM only IMO

PR #221, opened 2016-11-26
Drop old versions

greg0ire:
I am targetting this branch, because bumping minor version of some deps is considered major.

## Changelog

```markdown
### Removed
- support for PHP 5.3 through 5.5 was dropped
- support for Symfony 2.3 through 2.7 was dropped
```

## Todo
- [x] merge #222 
- [x] merge 3.x into master
- [x] add upgrade note
- [x] rebase

## Subject

This prepares the next major release

### DIFF
--- a/DependencyInjection/SonataFormatterExtension.php
+++ b/DependencyInjection/SonataFormatterExtension.php
@@ -58,7 +58,7 @@ class SonataFormatterExtension extends Extension
 
         if (!array_key_exists($config['default_formatter'], $config['formatters'])) {
             throw new \InvalidArgumentException(sprintf(
-                'SonataFormatterBundle - Invalid default formatter : %s, available : %s',
+                'SonataFormatterBundle - Invalid default formatter: %s, available: %s',
                 $config['default_formatter'],
                 json_encode(array_keys($config['formatters']))
             ));
--- a/DependencyInjection/SonataFormatterExtension.php
+++ b/DependencyInjection/SonataFormatterExtension.php
@@ -56,19 +56,12 @@ class SonataFormatterExtension extends Extension
             $loader->load('ckeditor.xml');
         }
 
-        // NEXT_MAJOR: remove this if block
-        if (!isset($config['default_formatter'])) {
-            @trigger_error(
-                'Not setting the default_formatter configuration node is deprecated since 3.x,'.
-                ' and will no longer be supported in 4.0.',
-                E_USER_DEPRECATED
-            );
-            reset($config['formatters']);
-            $config['default_formatter'] = key($config['formatters']);
-        }
-
         if (!array_key_exists($config['default_formatter'], $config['formatters'])) {
-            throw new \InvalidArgumentException(sprintf('SonataFormatterBundle - Invalid default formatter : %s, available : %s', $config['default_formatter'], json_encode(array_keys($config['formatters']))));
+            throw new \InvalidArgumentException(sprintf(
+                'SonataFormatterBundle - Invalid default formatter : %s, available : %s',
+                $config['default_formatter'],
+                json_encode(array_keys($config['formatters']))
+            ));
         }
 
         $pool = $container->getDefinition('sonata.formatter.pool');
@@ -78,7 +71,12 @@ class SonataFormatterExtension extends Extension
             if (count($configuration['extensions']) == 0) {
                 $env = null;
             } else {
-                $env = new Reference($this->createEnvironment($container, $code, $container->getDefinition($configuration['service']), $configuration['extensions']));
+                $env = new Reference($this->createEnvironment(
+                    $container,
+                    $code,
+                    $container->getDefinition($configuration['service']),
+                    $configuration['extensions']
+                ));
             }
 
             $pool->addMethodCall('add', array($code, new Reference($configuration['service']), $env));
--- a/Formatter/Pool.php
+++ b/Formatter/Pool.php
@@ -12,7 +12,7 @@
 namespace Sonata\FormatterBundle\Formatter;
 
 use Psr\Log\LoggerAwareInterface;
-use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerAwareTrait;
 use Psr\Log\NullLogger;
 use Twig_Environment;
 use Twig_Error_Syntax;
@@ -20,6 +20,8 @@ use Twig_Sandbox_SecurityError;
 
 class Pool implements LoggerAwareInterface
 {
+    use LoggerAwareTrait;
+
     /**
      * @var array
      */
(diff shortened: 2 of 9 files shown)


greg0ire:
@sagikazarmark can you please give this a review? I followed your `TODO` instructions.

greg0ire:
ping @sonata-project/contributors

Formatter/Pool.php
```diff
@@ -34,15 +34,13 @@ class Pool
     protected $logger;
 
     /**
-     * @param LoggerInterface|null $logger
      * @param string|null          $defaultFormatter
+     * @param LoggerInterface|null $logger
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter, LoggerInterface $logger = null)
```
core23: Please add an upgrade note for this.

sagikazarmark:
@greg0ire Sure, but I lost the context: what TODO instructions? 😄

sagikazarmark:
Ah, ok, found it

DependencyInjection/SonataFormatterExtension.php
```diff
@@ -56,25 +56,29 @@ public function load(array $configs, ContainerBuilder $container)
...
 
-        // TODO: To be removed when major version is changed
-        if (!isset($config['default_formatter'])) {
-            reset($config['formatters']);
-            $config['default_formatter'] = key($config['formatters']);
-        }
-
         if (!array_key_exists($config['default_formatter'], $config['formatters'])) {
-            throw new \InvalidArgumentException(sprintf('SonataFormatterBundle - Invalid default formatter : %s, available : %s', $config['default_formatter'], json_encode(array_keys($config['formatters']))));
+            throw new \InvalidArgumentException(sprintf(
```
sagikazarmark: I know you are against it, and I am not trying to start a flame war here again, but AFAIK the Symfony CS says exceptions should not be split into multiple lines. Am I wrong?

Formatter/Pool.php
sagikazarmark: Maybe also mention that passing null will no longer automatically make the first formatter the default one?

sagikazarmark: Actually null should be an invalid value in the first place.

DependencyInjection/SonataFormatterExtension.php
greg0ire: Fabien said it to me, but I am not sure they put it in their CS rules though (and I think the arguments he gave are not good enough, but can understand that he does not want to have to touch 10000 files over this). And anyway, there is a `sprintf` here, and we usually split `sprintf` calls, don't we?

Formatter/Pool.php
greg0ire: > Actually null should be an invalid value in the first place.

That's right. I think I'll have to make another PR on 3.x to fix that, and maybe also add deprecation notices for each of the `TODOs`

DependencyInjection/SonataFormatterExtension.php
sagikazarmark: As said, I saw quite a few rounds of debates about this and I am not up to date with the latest decisions on this front, I just wanted to point a possible discrepancy.

Formatter/Pool.php
sagikazarmark: Null is valid on the 3.x branch IMO, since it's optional. When it is null, the first one is automatically selected, this is necessary for BC. Otherwise I agree, deprecation warnings should be added. (Actually I am not sure why I didn't add any)

greg0ire: > Actually I am not sure why I didn't add any

Probably because this was quite old ;)

I'll remove the null in the signature then

DependencyInjection/SonataFormatterExtension.php
greg0ire: Personally, in this case I'll go as far as to prefer discrepancy, so as long as someone does not point me at some docs showing this is part of the sf coding standard, which I think is full of very questionable decisions like Yoda conditions or non-multiline-signatures that are 500 chars long (not kidding, there are things like this in sf), then I'll keep using and asking for linebroken exceptions. It gives better diffs, and people can actually read the exception messages that way.
I don't know why people are so afraid of hitting the carriage return key.

sagikazarmark: > but can understand that he does not want to have to touch 10000 files over this

That's the why

greg0ire: Sometimes I wonder if consistency is that important: if you do something wrong and hard to fix many times, you just give up and carve in your standards : "do it the wrong way"

Formatter/Pool.php
```diff
@@ -34,15 +34,13 @@ class Pool
...
-     * @param string|null          $defaultFormatter
+     * @param string          $defaultFormatter
+     * @param LoggerInterface $logger
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter, LoggerInterface $logger)
     {
-        $this->logger = $logger;
-
-        // TODO: This should become a required first parameter when the major version is changed
```
greg0ire: @sagikazarmark I think there is no point anymore to switch the arguments if the logger can't be null, am I right?

Formatter/Pool.php
greg0ire: Wait I got confused and removed null from the logger signature. My bad.

greg0ire: > Actually null should be an invalid value in the first place.

@sagikazarmark : called `cannotBeEmpty` to take care of that.

sagikazarmark:
I guess this needs to be rebased once 3.x is merged into master, right?

greg0ire:
Yes exactly. And reworked a bit also (because of the changes with the logger)

greg0ire:
@sagikazarmark it's ready, please review !

DependencyInjection/SonataFormatterExtension.php
```diff
@@ -56,19 +56,12 @@ public function load(array $configs, ContainerBuilder $container)
...
-                E_USER_DEPRECATED
-            );
-            reset($config['formatters']);
-            $config['default_formatter'] = key($config['formatters']);
-        }
-
         if (!array_key_exists($config['default_formatter'], $config['formatters'])) {
-            throw new \InvalidArgumentException(sprintf('SonataFormatterBundle - Invalid default formatter : %s, available : %s', $config['default_formatter'], json_encode(array_keys($config['formatters']))));
+            throw new \InvalidArgumentException(sprintf(
+                'SonataFormatterBundle - Invalid default formatter : %s, available : %s',
```
sagikazarmark: Hm, spaces before colons?

greg0ire:
> Pure beauty

:blush: thanks!

greg0ire:
@sonata-project/contributors , you already approved, but I changed many things, so maybe a final review before the merge?

sagikazarmark:
TBH I don't understand why approval is not automatically reset when more changes are added to the repo.

greg0ire:
> TBH I don't understand why approval is not automatically reset when more changes are added to the repo.

I think it would be hard to get right, because you'd have to separate pedantic changes from significant ones. For instance I just restored `UPGRADE-4.0.md` that I deleted by mistake, I don't think it warrants a reset.

sagikazarmark:
>  think it would be hard to get right, because you'd have to separate pedantic changes from significant ones.

Well, I would rather make sure all the time. The same applies to business: you don't deploy ANYTHING, not even a char change to production without approval.

I agree that it could actually be quite uncomfortable in some cases, but that's the only way to keep the formalities.

greg0ire:
I think it depends on the human resources you have on your team. Maybe Github will make this configurable someday?

Formatter/Pool.php
```diff
@@ -12,14 +12,16 @@
...
 
 use Psr\Log\LoggerAwareInterface;
-use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerAwareTrait;
 use Psr\Log\NullLogger;
 use Twig_Environment;
 use Twig_Error_Syntax;
 use Twig_Sandbox_SecurityError;
 
 class Pool implements LoggerAwareInterface
```
core23: Do we still need the interface when using the trait?

greg0ire: We inject the logger ourselves, so I'd say no… but if we want the class to benefit from some automatic logger injection we're not aware about in some framework, this might be necessary. Doesn't hurt IMO. @sagikazarmark , what do you think?

sagikazarmark: LoggerAware can be used as a marker that it requires a log, so I would say keep it. Doesn't hurt.

greg0ire: @core23 @OskarStark  please merge this if that's ok with you

core23:
Thanks @greg0ire

PR #222, opened 2016-12-03
Add missing deprecations

greg0ire:
I am targetting this branch, because this is BC (addind a deprecation notice).

## Changelog

```markdown
### Deprecated
- not specifying the `default_formatter` configuration node is deprecated
- specifying a logger through the `SonataFormatterBundle\Formatter\Pool` constructor is deprecated in favor of specifying it through `setLogger` method
- not specifying a default formatter to the `SonataFormatterBundle\Formatter\Pool` constructor
```

## Subject

See #221

### DIFF
--- a/DependencyInjection/SonataFormatterExtension.php
+++ b/DependencyInjection/SonataFormatterExtension.php
@@ -56,7 +56,7 @@ class SonataFormatterExtension extends Extension
             $loader->load('ckeditor.xml');
         }
 
-        // TODO: To be removed when major version is changed
+        // NEXT_MAJOR: remove this if block
         if (!isset($config['default_formatter'])) {
             reset($config['formatters']);
             $config['default_formatter'] = key($config['formatters']);
@@ -67,7 +67,7 @@ class SonataFormatterExtension extends Extension
         }
 
         $pool = $container->getDefinition('sonata.formatter.pool');
-        // TODO: This should become the first (zero-indexed) argument when the major version is changed
+        // NEXT_MAJOR: This should become the first (zero-indexed) argument
         $pool->addArgument($config['default_formatter']);
 
         foreach ($config['formatters'] as $code => $configuration) {
--- a/DependencyInjection/SonataFormatterExtension.php
+++ b/DependencyInjection/SonataFormatterExtension.php
@@ -72,7 +72,6 @@ class SonataFormatterExtension extends Extension
         }
 
         $pool = $container->getDefinition('sonata.formatter.pool');
-        // NEXT_MAJOR: This should become the first (zero-indexed) argument
         $pool->addArgument($config['default_formatter']);
 
         foreach ($config['formatters'] as $code => $configuration) {
--- a/DependencyInjection/SonataFormatterExtension.php
+++ b/DependencyInjection/SonataFormatterExtension.php
@@ -58,6 +58,11 @@ class SonataFormatterExtension extends Extension
 
         // NEXT_MAJOR: remove this if block
         if (!isset($config['default_formatter'])) {
+            @trigger_error(
+                'Not setting the default_formatter configuration node is deprecated since 3.x,'.
+                ' and will no longer be supported in 4.0.',
+                E_USER_DEPRECATED
+            );
             reset($config['formatters']);
             $config['default_formatter'] = key($config['formatters']);
         }
--- a/Formatter/Pool.php
+++ b/Formatter/Pool.php
@@ -140,11 +140,19 @@ class Pool implements LoggerAwareInterface
                 $text = $env->render($text);
             }
         } catch (Twig_Error_Syntax $e) {
-            $this->logger->critical(sprintf('[FormatterBundle::transform] %s - Error while parsing twig template : %s', $code, $e->getMessage()), array(
+            $this->logger->critical(sprintf(
+                '[FormatterBundle::transform] %s - Error while parsing twig template : %s',
+                $code,
+                $e->getMessage()
+            ), array(
                 'text' => $text,
             ));
         } catch (Twig_Sandbox_SecurityError $e) {
-            $this->logger->critical(sprintf('[FormatterBundle::transform] %s - the user try an non white-listed keyword : %s', $code, $e->getMessage()), array(
+            $this->logger->critical(sprintf(
+                '[FormatterBundle::transform] %s - the user try an non white-listed keyword : %s',
+                $code,
+                $e->getMessage()
+            ), array(
                 'text' => $text,
             ));
         }
--- a/Formatter/Pool.php
+++ b/Formatter/Pool.php
@@ -11,12 +11,14 @@
 
 namespace Sonata\FormatterBundle\Formatter;
 
+use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Twig_Environment;
 use Twig_Error_Syntax;
 use Twig_Sandbox_SecurityError;
 
-class Pool
+class Pool implements LoggerAwareInterface
 {
     /**
      * @var array
@@ -29,20 +31,60 @@ class Pool
     protected $defaultFormatter;
 
     /**
+     * NEXT_MAJOR: use LoggerAwareTrait.
+     *
      * @var LoggerInterface
      */
     protected $logger;
 
     /**
-     * @param LoggerInterface|null $logger
-     * @param string|null          $defaultFormatter
+     * @param string|null $defaultFormatter
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter = null)
+    {
+        if (func_num_args() == 2) {
+            list($logger, $defaultFormatter) = func_get_args();
+            $this->logger = $logger;
+            $this->defaultFormatter = $defaultFormatter;
+        } elseif (func_num_args() == 1) {
+            if ($defaultFormatter instanceof LoggerInterface) {
+                $this->logger = $defaultFormatter;
+            } else {
+                // NEXT_MAJOR: Only keep this block
+                $this->defaultFormatter = $defaultFormatter;
+            }
+        }
+
+        // NEXT_MAJOR: keep the else block only
+        if ($this->logger) {
+            @trigger_error(sprintf(
+                'Providing a logger to %s through the constructor is deprecated since 3.x'.
+                ' and will no longer be possible in 4.0'.
+                ' This argument should be provided through the setLogger() method.',
+                __CLASS__
+            ), E_USER_DEPRECATED);
+        } else {
+            $this->logger = new NullLogger();
+        }
+
+        // NEXT_MAJOR: make defaultFormatter required
+        if (is_null($this->defaultFormatter)) {
+            @trigger_error(sprintf(
+                'Not providing the defaultFormatter argument to %s is deprecated since 3.x.'.
+                ' This argument will become mandatory in 4.0.',
+                __METHOD__
+            ), E_USER_DEPRECATED);
+        }
+    }
+
+    /**
+     * NEXT_MAJOR: use Psr\Log\LoggerAwareTrait.
+     *
+     * @param LoggerInterface will be used to report errors
+     */
+    public function setLogger(LoggerInterface $logger)
     {
         $this->logger = $logger;
-
-        // NEXT_MAJOR: This should become a required first parameter
-        $this->defaultFormatter = $defaultFormatter;
     }
 
     /**
@@ -98,17 +140,13 @@ class Pool
                 $text = $env->render($text);
             }
         } catch (Twig_Error_Syntax $e) {
-            if ($this->logger) {
-                $this->logger->critical(sprintf('[FormatterBundle::transform] %s - Error while parsing twig template : %s', $code, $e->getMessage()), array(
-                    'text' => $text,
-                ));
-            }
+            $this->logger->critical(sprintf('[FormatterBundle::transform] %s - Error while parsing twig template : %s', $code, $e->getMessage()), array(
+                'text' => $text,
+            ));
         } catch (Twig_Sandbox_SecurityError $e) {
-            if ($this->logger) {
-                $this->logger->critical(sprintf('[FormatterBundle::transform] %s - the user try an non white-listed keyword : %s', $code, $e->getMessage()), array(
-                    'text' => $text,
-                ));
-            }
+            $this->logger->critical(sprintf('[FormatterBundle::transform] %s - the user try an non white-listed keyword : %s', $code, $e->getMessage()), array(
+                'text' => $text,
+            ));
         }
 
         return $text;
--- a/Formatter/Pool.php
+++ b/Formatter/Pool.php
@@ -41,7 +41,7 @@ class Pool
     {
         $this->logger = $logger;
 
-        // TODO: This should become a required first parameter when the major version is changed
+        // NEXT_MAJOR: This should become a required first parameter
         $this->defaultFormatter = $defaultFormatter;
     }
 
@@ -127,7 +127,7 @@ class Pool
      */
     public function getDefaultFormatter()
     {
-        // TODO: This should be removed when the major version is changed
+        // NEXT_MAJOR: This should be removed
         if (is_null($this->defaultFormatter)) {
             reset($this->formatters);
 
--- a/Formatter/Pool.php
+++ b/Formatter/Pool.php
@@ -146,6 +146,7 @@ class Pool implements LoggerAwareInterface
                 $e->getMessage()
             ), array(
                 'text' => $text,
+                'exception' => $e,
             ));
         } catch (Twig_Sandbox_SecurityError $e) {
             $this->logger->critical(sprintf(
@@ -154,6 +155,7 @@ class Pool implements LoggerAwareInterface
                 $e->getMessage()
             ), array(
                 'text' => $text,
+                'exception' => $e,
             ));
         }
 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -10,6 +10,48 @@ sonata_formatter:
     default_formatter: my_formatter
 ```
 
+## Deprecated injecting a logger through the Sonata\FormatterBundle\Formatter\Pool constructor
+
+Providing a logger to a `Sonata\FormatterBundle\Formatter\Pool` instance should
+now be done through the `setLogger` method.
+
+
+Before:
+
+```php
+use Sonata\FormatterBundle\Formatter\Pool;
+
+new Pool($myLogger, 'myFormatter');
+```
+
+After:
+
+```php
+$pool = new Pool('myFormatter');
+$pool->setLogger($myLogger);
+```
+
+## Deprecated unspecified default formatter in Sonata\FormatterBundle\Formatter\Pool constructor
+
+Instantiating the `Sonata\FormatterBundle\Formatter\Pool` class should be done
+with a `$defaultFormatter` argument.
+
+Before:
+
+```php
+use Sonata\FormatterBundle\Formatter\Pool;
+
+new Pool();
+```
+
+After:
+
+```php
+use Sonata\FormatterBundle\Formatter\Pool;
+
+new Pool('myFormatter');
+```
+
 UPGRADE FROM 3.0 to 3.1
 =======================
 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,15 @@
 UPGRADE 3.x
 ===========
 
+## Deprecated unspecified default_formatter configuration node
+
+The `default_formatter` configuration node will become required.
+
+```yaml
+sonata_formatter:
+    default_formatter: my_formatter
+```
+
 UPGRADE FROM 3.0 to 3.1
 =======================
 
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -32,7 +32,7 @@ class Configuration implements ConfigurationInterface
 
         $rootNode
             ->children()
-                ->scalarNode('default_formatter')->end() // TODO: make it required when the major version is changed
+                ->scalarNode('default_formatter')->end() // NEXT_MAJOR: make this required
                 ->arrayNode('formatters')
                     ->useAttributeAsKey('name')
                     ->prototype('array')
--- a/Resources/config/formatter.xml
+++ b/Resources/config/formatter.xml
@@ -8,7 +8,9 @@
     </parameters>
     <services>
         <service id="sonata.formatter.pool" class="Sonata\FormatterBundle\Formatter\Pool">
-            <argument type="service" id="logger"/>
+            <call method="setLogger">
+                <argument type="service" id="logger"/>
+            </call>
         </service>
         <service id="sonata.formatter.text.markdown" class="%sonata.formatter.text.markdown.class%">
             <tag name="sonata.text.formatter" code="markdown"/>
--- a/Tests/Form/EventListener/FormatterListenerTest.php
+++ b/Tests/Form/EventListener/FormatterListenerTest.php
@@ -21,7 +21,7 @@ class FormatterListenerTest extends \PHPUnit_Framework_TestCase
     {
         $this->setExpectedException('RuntimeException');
 
-        $pool = new Pool();
+        $pool = $this->getPool();
 
         $listener = new FormatterListener($pool, '[format]', '[source]', '[target]');
 
@@ -41,7 +41,7 @@ class FormatterListenerTest extends \PHPUnit_Framework_TestCase
             return strtoupper($text);
         }));
 
-        $pool = new Pool();
+        $pool = $this->getPool();
         $pool->add('myformat', $formatter);
 
         $listener = new FormatterListener($pool, '[format]', '[source]', '[target]');
@@ -61,5 +61,13 @@ class FormatterListenerTest extends \PHPUnit_Framework_TestCase
         );
 
         $this->assertSame($expected, $event->getData());
+    }
+
+    private function getPool()
+    {
+        $pool = new Pool('whatever');
+        $pool->setLogger($this->getMock('Psr\Log\LoggerInterface'));
+
+        return $pool;
     }
 }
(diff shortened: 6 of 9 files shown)


greg0ire:
Please review @sagikazarmark

DependencyInjection/SonataFormatterExtension.php
```diff
@@ -56,8 +56,13 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('ckeditor.xml');
         }
 
-        // TODO: To be removed when major version is changed
+        // NEXT_MAJOR: remove this if block
         if (!isset($config['default_formatter'])) {
+            @trigger_error(<<<'EOT'
+Not setting the default_formatter configuration node is deprecated since 3.x,
```
sagikazarmark: Well, I don't really like this kind of line splitting within strings. Does not make things easier to read IMO.

sagikazarmark: Also, maybe add error triggering in the pool as well?

greg0ire: > Well, I don't really like this kind of line splitting within strings. Does not make things easier to read IMO.

I could argue that it is a matter of taste, but is it? I mean you did not have to scroll horizontally to read this piece of text, so it really was easier to read. If I didn't, the line would be ~ 130 chars wide because of indention.

> Also, maybe add error triggering in the pool as well?

I thought about it, but how would you achieve that? Because you can't trigger a deprecation if you don't offer a new way. Would you drop the type hinting and do duck typing? It's possible but a bit ugly. Another possiblity I thought about: make the logger mandatory too (people that do not want to log can use a null logger). If we do that, we are no longer forced to switch arguments. What do you think?

greg0ire: I'm gonna go ahead and do it while I have some spare time.

greg0ire: @sagikazarmark : I pushed a new version 2bab6fd7786a48ddfbdb44c24c1bec5cf7b56e4d with null logger deprecated, tell me what you think.

greg0ire: ping @sagikazarmark ?

greg0ire: No response. @sonata-project/contributors please review.

core23: Using Nowdoc for this short text is to excessive. 

Same for the other code fragment.

greg0ire: What would you do ? Concatenation? All on one line? I'll try concatenation if everyone is put off by this.

@sagikazarmark :

> Well, I don't really like this kind of line splitting within strings. Does not make things easier to read IMO.

By that, did you mean the string the code, or the produced string? I think I might have misunderstood this.

greg0ire:
@sagikazarmark @core23 here is version without splitting the produced string. I don't like it because I have to be careful to add the space on line 63 if I want to produce a nice string, and because it's cumbersome to create and edit (if I have to add something in the middle of the string, I have to move the point where I use concatenation). Also, notice how this would go beyond the 120 chars soft limit and make people scroll to review if I didn't use concatenation. So no, I don't think nowdoc is excessive here, but I listen when both of you say you do not like it.

sagikazarmark:
For me enforcing the char limit in every case doesn't sound like a good idea. Keep in mind that those messages are for logs or error pages, not for the code itself. Also, splitting arguments into multiple lines does not add ANY overhead, the concatenation does, even if it is negligible, but it's there for the sake of char limiting? I am definitely against it.

Formatter/Pool.php
```diff
@@ -39,10 +39,19 @@ class Pool
      */
     public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
     {
+        // NEXT_MAJOR: make both parameters required
```
sagikazarmark: Do we really need a logger here? Personally I would prefer leaving it optional and using the NullLogger as a fallback so that there is no need to check the existence of the logger. I would even consider removing it from the constructor and using the LoggerAwareInterface.

greg0ire:
> Keep in mind that those messages are for logs or error pages, not for the code itself. 

That's their main purpose indeed, but the message should also be reviewed, and might be edited, or just read by someone who wished to understand what the condition he just read is about.

> Also, splitting arguments into multiple lines does not add ANY overhead, the concatenation does, even if it is negligible, but it's there for the sake of char limiting? I am definitely against it.

I have zero concern about overhead here, that would be a typical case of premature performance optimization. What I want to optimize for here is readability and maintainability. I want to avoid things you can see in Symfony like[this kind message I linked to earlier where you have to scroll again and again](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/MessageSelector.php#L83), and things you can see in Doctrine, [like this kind of unmaintainable message](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/ORMException.php#L70). Heredoc was IMO made precisely for this kind of case, and if you start not using it with short cases, you can very well get carried away with long strings like in Symfony or Doctrine. That's what I want to avoid here.

I can see though how heredoc / nowdoc have their own problems : 
- it breaks the indention;
- the delimiters are long to write, and look ugly.

I'd like to promote them however, because the advantages show up quickly, once you go past parsing the delimiters, the enclosed message is very readable, and in a very objective and justifiable manner : 
- no concatenation in the middle of the string, which makes it easy to edit, and, arguably, read;
- no need to scroll, which makes it indubitably easier to review;
- you can use both single and double quotes without any escaping, which makes it easier to write.

Your main point might be "one log should be equal to one line". I tend to instinctively agree with that, but can't say why. I mean, what are you going to do with that, count the number of logs? Or is it about parsing logs with ease? Anyway, it this rule is so important, shouldn't it be the concern of the logging library, which could easily strip carriage returns and extra whitespace?

Formatter/Pool.php
greg0ire: So I thought about it, and I think BC might be actually achievable with some `func_num_args` magic. Let me give it a try.

sagikazarmark:
I think all arguments are given, it's up to the maintainers to decide.

greg0ire:
> I think all arguments are given, it's up to the maintainers to decide.

Indeed. I'm working on the LoggerAwareInterface implementation right now, almost finished. Not using heredoc because I don't want this discussion to hold this PR back.

Formatter/Pool.php
```diff
@@ -34,15 +36,43 @@ class Pool
...
+        }
         $this->defaultFormatter = $defaultFormatter;
+        // NEXT_MAJOR: make defaultFormatter required
+        if (is_null($defaultFormatter)) {
+            @trigger_error(sprintf(
+                'Not providing the defaultFormatter argument to %s is deprecated since 3.x.'.
+                ' This argument will become mandatory in 4.0.',
+                __METHOD__
+            ), E_USER_DEPRECATED);
+        }
```
greg0ire: @sagikazarmark : I'm done with the `LoggerAwareInterface` implementation (I think). Looks good?

Formatter/Pool.php
```diff
@@ -34,15 +36,43 @@ class Pool
     protected $logger;
```
sagikazarmark: If you use the trait from the psr/log package you can drop this

Formatter/Pool.php
```diff
@@ -34,15 +36,43 @@ class Pool
...
+        if (is_null($defaultFormatter)) {
+            @trigger_error(sprintf(
+                'Not providing the defaultFormatter argument to %s is deprecated since 3.x.'.
+                ' This argument will become mandatory in 4.0.',
+                __METHOD__
+            ), E_USER_DEPRECATED);
+        }
+    }
+
+    final public function setLogger(LoggerInterface $logger)
```
sagikazarmark: This as well

Formatter/Pool.php
```diff
@@ -136,4 +172,17 @@ public function getDefaultFormatter()
...
+     * Falls back to a NullLogger so that we do not have to check for the logger
+     * before calling it.
+     */
+    private function getLogger()
+    {
+        if ($this->logger) {
+            return $this->logger;
+        }
+
+        return new NullLogger();
```
sagikazarmark: Hm, not sure about lazy initialization. I tend to avoid it. Also, you didn't spare anything, the same condition is still there. I would probably set a null logger in the constructor by default.

greg0ire: Nice suggestion, I'll do that.

Formatter/Pool.php
greg0ire: will do 👍

greg0ire:
@sagikazarmark : all good now?

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+     * @param string|null $defaultFormatter
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter = null)
     {
-        $this->logger = $logger;
-
-        // TODO: This should become a required first parameter when the major version is changed
+        $this->logger = new NullLogger();
+        // NEXT_MAJOR: remove this block
```
sagikazarmark: Add a new line here?

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+        if (func_num_args() == 1 && $defaultFormatter instanceof LoggerInterface) {
+            $this->logger = $defaultFormatter;
+            @trigger_error(sprintf(
+                'Providing a logger to %s through the constructor is deprecated since 3.x'.
+                ' and we no longer be possible in 4.0'.
+                ' This argument should be provided through the setLogger() method.',
+                __METHOD__
+            ), E_USER_DEPRECATED);
+
+            return;
```
sagikazarmark: Returning here is not a good idea as the default formatter is null and we should trigger a deprecation error. Just set the defaultFormatter to null and let the rest of the logic run.

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+        }
```
sagikazarmark: Add a newline here?

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+        $this->logger = new NullLogger();
+        // NEXT_MAJOR: remove this block
+        if (func_num_args() == 2) {
+            list($logger, $defaultFormatter) = func_get_args();
+            $this->logger = $logger;
+            $this->defaultFormatter = $defaultFormatter;
+
+            return;
+        }
+        if (func_num_args() == 1 && $defaultFormatter instanceof LoggerInterface) {
```
sagikazarmark: Elseif? No need to evaluate this condition if there are two arguments.

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+            $this->logger = $defaultFormatter;
+            @trigger_error(sprintf(
```
sagikazarmark: Shouldn't this error be triggered if there are two arguments too?

sagikazarmark: Maybe: if this->logger is set then trigger error else this->logger = new NullLogger().

Formatter/Pool.php
```diff
@@ -29,20 +31,56 @@ class Pool
...
+            ), E_USER_DEPRECATED);
+        }
+    }
+
+    /**
+     * NEXT_MAJOR: use Psr\Log\LoggerAwareTrait.
+     *
+     * @param LoggerInterface will be used to report errors
+     */
+    final public function setLogger(LoggerInterface $logger)
```
sagikazarmark: Not sure how this works in the interface/trait, but I would use the exact same signature to avoid breaking BC (though removing final is not a BC breaking operation). As far as I remember the trait exposes a fluent interface. I might be wrong, but if not, this should be fluent as well.

greg0ire: The interface is not fluent, and as you said, removing final is not a BC-break, but there is little point to making it final if we are sure it won't be in the future.

greg0ire:
@sagikazarmark : thanks for your review, I reworked the constructor entirely with your remarks in mind, hopefully the logic is better this time.

Formatter/Pool.php
```diff
@@ -29,20 +31,59 @@ class Pool
...
+                'Providing a logger to %s through the constructor is deprecated since 3.x'.
+                ' and we no longer be possible in 4.0'.
+                ' This argument should be provided through the setLogger() method.',
+                __METHOD__
+            ), E_USER_DEPRECATED);
+        } else {
+            $this->logger = new NullLogger();
+        }
+        // NEXT_MAJOR: make defaultFormatter required
+        if (is_null($defaultFormatter)) {
```
sagikazarmark: This issue is still not resolved: if the first and only arg is a logger, default formatter won't be null here, missing the deprecation trigger. Either check the class property, or set the default formatter on line 51 to null.

greg0ire: Oh right, I missed that, sorry! I'll do the latter.

greg0ire:
@sagikazarmark : fixed the default formatter check

Formatter/Pool.php
```diff
@@ -29,20 +31,60 @@ class Pool
...
     /**
-     * @param LoggerInterface|null $logger
-     * @param string|null          $defaultFormatter
+     * @param string|null $defaultFormatter
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter = null)
     {
-        $this->logger = $logger;
+        if (func_num_args() == 2) {
```
sagikazarmark: NEXT_MAJOR: Remove this block

Formatter/Pool.php
```diff
@@ -29,20 +31,60 @@ class Pool
...
+                $this->defaultFormatter = $defaultFormatter;
+            }
+        }
 
-        // TODO: This should become a required first parameter when the major version is changed
-        $this->defaultFormatter = $defaultFormatter;
+        // NEXT_MAJOR: keep the else block only
+        if ($this->logger) {
+            @trigger_error(sprintf(
+                'Providing a logger to %s through the constructor is deprecated since 3.x'.
```
sagikazarmark: This will become "Providing a logger to Pool::__construct through the constructor...". Is this intended?

sagikazarmark: This is where the string split kicks in as well: if you change the message, you need to rearrange the strings as well.

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+After:
+
+```
+$pool = new Pool('myFormatter');
+$pool->setLogger($myLogger);
+```
+
+## Deprecated unspecified Sonata\FormatterBundle\Formatter\Pool defaultFormatter
+
+Instanciating the `Sonata\FormatterBundle\Formatter\Pool` class should be done
```
sagikazarmark: Instantiating

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+
+```yaml
+sonata_formatter:
+    default_formatter: my_formatter
+```
+
+## Deprecated specifying a logger in the Sonata\FormatterBundle\Formatter\Pool constructor
+
+Providing a logger to a `Sonata\FormatterBundle\Formatter\Pool` instance is
+optional, and optional dependencies should be provided through a setter.
```
sagikazarmark: IMO this is arguable, I would just stick to facts: Logger can now be injected via setter instead of passing it to the constructor.

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+## Deprecated unspecified default_formatter configuration node
+
+The `default_formatter` configuration node will become required.
+
+```yaml
+sonata_formatter:
+    default_formatter: my_formatter
+```
+
+## Deprecated specifying a logger in the Sonata\FormatterBundle\Formatter\Pool constructor
```
sagikazarmark: Maybe say injecting?

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+
+Providing a logger to a `Sonata\FormatterBundle\Formatter\Pool` instance is
+optional, and optional dependencies should be provided through a setter.
+
+Before:
+
+```
```
sagikazarmark: Missing php?

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+
+After:
+
+```
+$pool = new Pool('myFormatter');
+$pool->setLogger($myLogger);
+```
+
+## Deprecated unspecified Sonata\FormatterBundle\Formatter\Pool defaultFormatter
```
sagikazarmark: Deprecated unspecified default formatter in Sonata\FormatterBundle\Formatter\Pool constructor?

UPGRADE-3.x.md
```diff
@@ -1,6 +1,56 @@
...
+
+Instanciating the `Sonata\FormatterBundle\Formatter\Pool` class should be done
+with both a `$defaultFormatter` argument.
```
sagikazarmark: -both

sagikazarmark:
Added a few more comments, but the code looks ok to me.

Formatter/Pool.php
greg0ire: I moved the comment to line 34, to explain that line 35 must be kept, which would be harder to convey from line 45.

Formatter/Pool.php
greg0ire: Oh right, changing it to `__CLASS__`

> This is where the string split kicks in as well: if you change the message, you need to rearrange the strings as well.

Yeah but if you used nowdoc it's a breeze: you join the lines together again, hit `gqq` and you're done. If you use concatenation… that's another story.

UPGRADE-3.x.md
greg0ire: That's better, yes 👍

UPGRADE-3.x.md
greg0ire: One of my coworkers just made this mistake this week! Now I'm contaminated!

UPGRADE-3.x.md
greg0ire: Good catch!

greg0ire:
> Added a few more comments, but the code looks ok to me.

Hurray! @sonata-project/contributors , can you please give this a final review?

Formatter/Pool.php
```diff
@@ -11,12 +11,14 @@
...
 
+use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Twig_Environment;
 use Twig_Error_Syntax;
 use Twig_Sandbox_SecurityError;
 
-class Pool
+class Pool implements LoggerAwareInterface
```
OskarStark: this is per definition a BC break IMO

Formatter/Pool.php
```diff
@@ -29,20 +31,60 @@ class Pool
...
      */
-    public function __construct(LoggerInterface $logger = null, $defaultFormatter = null)
+    public function __construct($defaultFormatter = null)
     {
-        $this->logger = $logger;
+        if (func_num_args() == 2) {
+            list($logger, $defaultFormatter) = func_get_args();
+            $this->logger = $logger;
+            $this->defaultFormatter = $defaultFormatter;
+        } elseif (func_num_args() == 1) {
```
OskarStark: yoda please

Formatter/Pool.php
OskarStark: yoda please

Formatter/Pool.php
```diff
@@ -29,20 +31,60 @@ class Pool
...
+            }
+        }
 
-        // TODO: This should become a required first parameter when the major version is changed
-        $this->defaultFormatter = $defaultFormatter;
+        // NEXT_MAJOR: keep the else block only
+        if ($this->logger) {
+            @trigger_error(sprintf(
+                'Providing a logger to %s through the constructor is deprecated since 3.x'.
+                ' and we no longer be possible in 4.0'.
```
OskarStark: we or will?

Formatter/Pool.php
greg0ire: I don't think so. Please tell me how it would be one.

Formatter/Pool.php
greg0ire: will :P

Formatter/Pool.php
greg0ire: You can't assign a value to a function call. So this would protect us from nothing, and make things less readable.

Formatter/Pool.php
greg0ire: @OskarStark fixed! Good catch!

Formatter/Pool.php
sagikazarmark: Just wanted to say that. Yoda for function call?????

Formatter/Pool.php
sagikazarmark: I am curious about the reasoning as well. 😄

Formatter/Pool.php
greg0ire: @sagikazarmark : I always said the Jedi council has too much influence :) [Look how carried away they can get](https://github.com/symfony/symfony-docs/pull/5381)

sagikazarmark: Well, for me THIS is the point where sanity should be superior to consistency.

Formatter/Pool.php
greg0ire: Scheduled for next major

Formatter/Pool.php
OskarStark: thats why i said per definition, but if we fullfill the interface, everything is fine.

greg0ire:
Thanks @OskarStark , @core23, final review and merge? Or maybe we can merge like that since @sagikazarmark  has approved?

sagikazarmark:
I still have faith that @core23 will veto the string splitting. :trollface:

Formatter/Pool.php
sagikazarmark: Just for the theoretical debate: how is it a BC break per definition? If you don't actually implement the interface your code will be broken.

greg0ire:
> I still have faith that @core23 will veto the string splitting. :trollface:

Noooooo! Please don't do that! I already abandoned nowdoc because you asked, please don't coerce me into commiting any other atrocity.

@sagikazarmark joking aside, I'd like to know : 
1. is [this](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Translation/MessageSelector.php#L83) really ok for you? 
2. If yes, do you really think @OskarStark would have caught [this](https://github.com/sonata-project/SonataFormatterBundle/pull/222#discussion_r91902906) hadn't it be split?
3. if not :
    1. if not what is an acceptable char limit for you?
    2. would you always do [this](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/ORMException.php#L70) or is there also a limit where you would switch to nowdoc?

sagikazarmark:
Whoa, too much questions and logical links between them.

I think I already expressed that exception messages are NOT primarily for developers to read, while code needs to be read, so char limit makes sense there.

I also said that any kind of string magic adds an overhead, even if negligible. Anything like that JUST FOR readability is totally unacceptable IMO.

Most IDEs break the long lines automatically anyway, so the readability issue is not an issue then.

Char limit is not for string literals, it's for code. How do you apply char limit to a JSON document for example? The only reason you can do it here is that you can do it programatically. GOTO two points above

In the linked doctrine case I would probably use sprintf, since they mostly use concatenation because of the parameters in the string.

This is also a thing where sanity should be over consistency: you cannot apply a char limit anywhere, or you can, but it comes with a price which I don't want to pay.

I think enough reasons and arguments said, while on the other side I only see blindly applying char limit (no offense) and bringing the readability argument which I IMO is not a valid one (see reasoning above).

greg0ire:
> Whoa, too much questions and logical links between them.

I numbered them!

> NOT primarily for developers

Primarily. Still, it can be helpful for them. In our case, I think it clarifies a lot of things. So yes, char limit makes sense there too IMO, and makes sense in any file that is considered code (as opposed to data). But maybe that's where the argument boils down to? You view exception messages as pure data, while I think of them more as metadata associated with a piece of code.

> How do you apply char limit to a JSON document for example

I don't store JSON documents in code directly… that's data, not code.

> I also said that any kind of string magic adds an overhead, even if negligible.

Nowdoc doesn't add any, does it?

> I think enough reasons and arguments said, while on the other side I only see blindly applying char limit (no offense) and bringing the readability argument which I IMO is not a valid one (see reasoning above).

No offense taken, I'm happy with this dicussion, and hope you are too. I know it's a very pedantic debate, but it affects a lot of code.

sagikazarmark:
Well, that's why I said it's a matter of priorities and personal taste and should be up to the maintainers to decide. I don't actually think that any of our reasoning can be superior to each other. One thing I would keep in mind: consistency with the rest of the code base is important too, so if you want to raise this issue, I would be happier with a general one for all the projects.

greg0ire:
> consistency with the rest of the code base is important too, so if you want to raise this issue, I would be happier with a general one for all the projects.

I'll make sure to do that and ping you when I do. In the meantime, I think we can merge this as is b/c we already use concatenation and heredoc in Sonata for that purpose (guess who did this).

sagikazarmark:
> (guess who did this)

Bah, abominations slip in unnoticed is a downside of open source projects. :trollface:

DependencyInjection/SonataFormatterExtension.php
```diff
@@ -56,8 +56,13 @@ public function load(array $configs, ContainerBuilder $container)
             $loader->load('ckeditor.xml');
         }
 
-        // TODO: To be removed when major version is changed
+        // NEXT_MAJOR: remove this if block
         if (!isset($config['default_formatter'])) {
+            @trigger_error(
+                'Not setting the default_formatter configuration node is deprecated since 3.x,'.
```
core23: I wouldn't split this string in two parts just for the character limit, because searching for this particular string in the whole projects wouldn't work later. 

This is a general discussion I don't want to start :trollface:

sagikazarmark: How diplomatic.... 😝

greg0ire: > This is a general discussion I don't want to start :trollface:

Well why are you starting it then? All jokes aside, you both aprroved, so I'll just merge. But stay tuned for a real discussion about this, I'll make ~a blog post~ an issue soon.

sagikazarmark:
https://twitter.com/sagikazarmark/status/808757703250219008

greg0ire:
See https://github.com/sonata-project/dev-kit/issues/218

PR #228, opened 2017-01-21
Fixed rawhtml format in sonata_simple_formatter_type_widget

tbmatuka:
I am targetting 3.x, because it's just a small BC template fix.

## Changelog

```markdown
### Fixed
- Broken markitup header for rawhtml format in sonata_simple_formatter_type_widget
```

## Subject

When using rawhtml format markitup needs a "html" class to be added to the parent in order to show the header icons. This was already set up for sonata_formatter_type_widget, but not for sonata_simple_formatter_type_widget, so I fixed that. It's ugly but it works.

### DIFF
--- a/Resources/views/Form/formatter.html.twig
+++ b/Resources/views/Form/formatter.html.twig
@@ -90,6 +90,9 @@
 
             if (parent) {
                 parent.addClass('{{ format }}');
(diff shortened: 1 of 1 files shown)


greg0ire:
Thanks @tbmatuka

PR #234, opened 2017-02-17
Removed deprecations

jordisala1991:
I am targetting this branch, because this is BC.

## Subject

Removed some deprecations throwed by tests. There are two remaining but not part of this bundle. They are part from block bundle, and they are already fixed (not on a release tho).

### DIFF
--- a/Block/FormatterBlockService.php
+++ b/Block/FormatterBlockService.php
@@ -12,8 +12,8 @@
 namespace Sonata\FormatterBundle\Block;
 
 use Sonata\AdminBundle\Form\FormMapper;
-use Sonata\BlockBundle\Block\BaseBlockService;
 use Sonata\BlockBundle\Block\BlockContextInterface;
+use Sonata\BlockBundle\Block\Service\AbstractBlockService;
 use Sonata\BlockBundle\Model\BlockInterface;
 use Sonata\CoreBundle\Model\Metadata;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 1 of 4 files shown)


Block/FormatterBlockService.php
```diff
@@ -23,7 +23,7 @@
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
  */
-class FormatterBlockService extends BaseBlockService
+class FormatterBlockService extends AbstractBlockService
```
greg0ire: No need to change the version constraint b/c of that?

jordisala1991: Ugh, maybe, let me see...

jordisala1991: You're right. changed.

greg0ire:
Thanks @jordisala1991

Block/FormatterBlockService.php
TheCelavi: Ok, this change (inheritance) broke usage of FormatterBlockService in Sonata sandbox when composing the page. 

It can not be seen in demo, since last commit is (as I can see) from December 8th 2016. https://github.com/sonata-project/sandbox/commits/2.4-develop 

However, BaseBlockService does not implements use Sonata\BlockBundle\Block\Service\AdminBlockServiceInterface and therefore it can not be used in Sonata page bundle in admin as drag&drop control.

Is this by design? What is alternative? Hope you didn't intended to remove this feature...

TheCelavi: @jordisala1991 Is this a typo? Did you mean to extend `Sonata\BlockBundle\Block\Service\AbstractAdminBlockService` instead?

jordisala1991: https://github.com/sonata-project/SonataBlockBundle/blob/3.x/Block/BaseBlockService.php

This is funny, the deprecate message is telling one service, but the class is actually extending another, maybe this change was wrong, will check! Thanks @TheCelavi can you provide a PR? Also we should change deprecation message, because it is missleading...

TheCelavi: Yes, of course.

PR #241, opened 2017-03-11
Added compatibility with 4.0 of admin bundle and media bundle

jordisala1991:
I am targeting this branch, because this is BC and allows travis to pass.

Closes https://github.com/sonata-project/SonataMediaBundle/pull/1225

## Changelog

```markdown
### Added
- Added compatibility with 4.0 of admin bundle and media bundle
```
## Subject

We are blocking installs of dev-master of media-bundle and admin-bundle, this makes builds to fail when using those versions. This change is inspired on: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/composer.json#L73.

But actually on that same composer.json there is another dependency blocking 4.0.

How do we handle this? @sonata-project/contributors

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,8 +40,8 @@
         "sonata-project/media-bundle": "For using the admin media browser."
     },
     "conflict": {
-        "sonata-project/admin-bundle": "<3.0 || >= 4.0",
-        "sonata-project/media-bundle": "<3.0 || >= 4.0"
+        "sonata-project/admin-bundle": "<3.0",
+        "sonata-project/media-bundle": "<3.0"
     },
     "autoload": {
         "psr-4": { "Sonata\\FormatterBundle\\": "" },


greg0ire:
IMO we shouldn't test with master versions : what good does it bring?

jordisala1991:
It is not about testing with them. 

This PR focus on `MediaBundle dev-master` not being able to find an installable SonataFormatter bundle due to this restriction I am changing.

----

Anyway, one example could be SonataUserBundle. If we don't manage to make BC between FOSUser 1.x and 2.x, we will do the change directly on `master 4.x`. And people (me included) may wanto to install `SonataUser dev-master` (in order to use FOSUser 2 and SF 3) but keep the other SonataBundles on a `stable 3.x` tag.

greg0ire:
For SonataUserBundle, the good way would be to create a special branch for that, which we did.

greg0ire:
Indeed for the installation, I got confused and get it now.

composer.json
```diff
@@ -40,8 +40,8 @@
         "sonata-project/media-bundle": "For using the admin media browser."
     },
     "conflict": {
-        "sonata-project/admin-bundle": "<3.0 || >= 4.0",
-        "sonata-project/media-bundle": "<3.0 || >= 4.0"
+        "sonata-project/admin-bundle": "<3.0 || >=5.0",
```
core23: IMHO we should remove the upper limit if there is no special reason

jordisala1991: Yep, that's why I wanted to do, but I wasn't sure about it.

Do we change this?

jordisala1991: For me it is better because it will avoid troubles like the one we are trying to solve here and if there is a need in the future for the restriction we can add it then. WDYT @greg0ire ?

greg0ire: I agree with you, IMO we are misusing `conflict` here.

jordisala1991: If we all agree... then.. changed!

core23:
Thanks @jordisala1991

PR #242, opened 2017-03-11
Add compatibility with egeloen/ckeditor-bundle 5.0

jordisala1991:

## Changelog

```markdown
### Added
- Add compatibility with `egeloen/ckeditor-bundle 5.0`
```
## Subject

IvoryCkEditorBundle 5.0.0 was released some days ago, this PR enables to install it with SonataFormatterBundle

### DIFF
--- a/Tests/Form/Type/SimpleFormatterTypeTest.php
+++ b/Tests/Form/Type/SimpleFormatterTypeTest.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


core23:
No other changes necessary?

jordisala1991:
@core23 as far as the UPGRADE guide tells, there is not change required on our end: https://github.com/egeloen/IvoryCKEditorBundle/blob/master/UPGRADE.md
We do not instantiate any `CKEditorType`, we do not use those removed functions and I couldn't find any reference to `RouteAbsolute` (because it was changed to RouteType.

@greg0ire the coverage looks half good: we are using Ivory classes on:

FormatterType -> 67% coverage (there are few ifs missing, and a function missing also from coverage)
SimpleFormatterType -> 0% (it looks like a simpler version of the other looking at the code)

IMO it is fine, but if you want I could add test for SimpleFormatter and increase coverage for future changes.

jordisala1991:
Added Tests... and coverage decreased 26%, nice...

It is because phpunit.xml.dist was not properly configured

OskarStark:
Thank you @jordisala1991 👍

PR #256, opened 2017-06-28
remove if statement from twig loader class

piszczek:


greg0ire:
Kindly fill the PR template and ping us when you think it is OK.

PR #252, opened 2017-05-20
Fixed twig 2 bug when updating rich text

core23:
I am targeting this branch, because this is a bugfix.

Closes #248

## Changelog

```markdown
### Fixed
 - Fixed twig 2 bug when updating rich text
```

## Subject

This fix will finally support twig 2.

PR #262, opened 2017-08-11
Fixed passing null to transformer

core23:
I am targeting this branch, because this is a bugfix.

Follow up of #252

## Changelog

```markdown
### Fixed
 - Fixed passing null to transformer
```

## Subject

When using the formatter and allow null values, twig 2 couldn't handle null values and will exit with an exception:

```
Template "__string_template__XXX" is not defined.
```


OskarStark:
Thank you @core23 👍

PR #277, opened 2017-10-15
Removed symfony BC layer for old symfony versions

core23:

## Changelog

```markdown
### Removed
 - Removed BC layer for older symfony versions
```

### DIFF
--- a/docs/reference/ckeditor_medias.rst
+++ b/docs/reference/ckeditor_medias.rst
@@ -80,11 +80,11 @@ Here is an example to alter `shortDescription` field of the `ProductAdmin`:
 .. code-block:: php
 
     <?php
-    // BaseProductProvider.php
+    use Sonata\FormatterBundle\Form\Type\FormatterType;
 
     // ...
 
-    $formMapper->add('shortDescription', 'sonata_formatter_type', array(
+    $formMapper->add('shortDescription', FormatterType::class, array(
         'source_field'         => 'rawDescription',
         'source_field_options' => array('attr' => array('class' => 'span10', 'rows' => 20)),
         'format_field'         => 'descriptionFormatter',
@@ -127,6 +127,7 @@ Alternatively you can specify custom return image format per field:
 .. code-block:: php
 
     <?php
+    use Sonata\FormatterBundle\Form\Type\SimpleFormatterType;
 
     // ...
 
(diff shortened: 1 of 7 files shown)


Form/Type/FormatterType.php
```diff
@@ -212,11 +215,6 @@ public function configureOptions(OptionsResolver $resolver)
             'choices' => $formatters,
         ];
 
-        // NEXT_MAJOR: Remove the method_exists hack when dropping support for symfony < 2.7
-        if (count($formatters) > 1 && method_exists('Symfony\Component\Form\AbstractType', 'configureOptions')) {
```
jordisala1991: The first part of the if is no longer needed? NEXT_MAJOR comment only talks about removing `method_exists`

Form/Type/FormatterType.php
```diff
@@ -212,11 +215,6 @@ public function configureOptions(OptionsResolver $resolver)
             'choices' => $formatters,
```
jordisala1991: This should get moved to the `format_field_options`

Form/Type/FormatterType.php
core23: Don't know why we should only add this option when we have at least one formatter.

jordisala1991: https://github.com/sonata-project/SonataFormatterBundle/pull/211#issuecomment-240460367

Here is the discussion, not sure if it applies now, but seems like when we have one formatter, is displayed as a hidden input, and on SF 2.8 atleast it does not have `choice_translation_domain`

core23: Fixed

jordisala1991:
Triggering build again after test fixing.

core23:
Please review again @jordisala1991

docs/reference/ckeditor_medias.rst
```diff
@@ -84,7 +84,7 @@ Here is an example to alter `shortDescription` field of the `ProductAdmin`:
 
     // ...
 
-    $formMapper->add('shortDescription', 'sonata_formatter_type', array(
+    $formMapper->add('shortDescription', FormatterType::class, array(
```
greg0ire: Please add use statements

core23: Good catch!

PR #279, opened 2017-10-17
Fix twig 2.x deprecated use of _self

loconox:
I am targeting this branch, because it's a bug fix.

## Changelog

```markdown
### Fixed
- Fix the deprecated use of _self in twig template
```
## Subject

In twig 2.x, `_self` returns the template name and need to be fix in this view.

### DIFF
--- a/Resources/views/Ckeditor/browser.html.twig
+++ b/Resources/views/Ckeditor/browser.html.twig
@@ -46,6 +46,7 @@ file that was distributed with this source code.
 
 {% import _self as tree %}
 {% macro navigate_child(collection, admin, root, current_category, depth) %}
+    {% import _self as tree %}
     {% if root and collection|length == 0 %}
         <div>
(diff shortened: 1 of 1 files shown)


greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

> In twig 2.x, _self returns the template name and need to be fix in this view.

Would be great to have that kind of comment /  a source in the commit message for instance

greg0ire:
Perfect job with the commit message @loconox !

PR #300, opened 2017-12-04
Use FormRenderer runtime to maintain compatibility with Symfony 3.4

jordisala1991:

## Changelog

```markdown
### Fixed
- Fix for getRuntime on Symfony older than 3.4
```

## To do

- [x] Update the tests

## Subject

TwigRenderer is used as runtime until SF 3.4.
This code makes sure we retrieve the correct runtime for SF < 3.4

### DIFF
--- a/src/Controller/CkeditorAdminController.php
+++ b/src/Controller/CkeditorAdminController.php
@@ -13,6 +13,11 @@ namespace Sonata\FormatterBundle\Controller;
 
 use Sonata\MediaBundle\Controller\MediaAdminController;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
+use Symfony\Bridge\Twig\AppVariable;
+use Symfony\Bridge\Twig\Command\DebugCommand;
+use Symfony\Bridge\Twig\Extension\FormExtension;
+use Symfony\Bridge\Twig\Form\TwigRenderer;
+use Symfony\Component\Form\FormRenderer;
 use Symfony\Component\Form\FormView;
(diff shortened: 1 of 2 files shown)


src/Controller/CkeditorAdminController.php
```diff
@@ -187,12 +192,19 @@ private function setFormTheme(FormView $formView, $theme)
...
             return;
         }
-        $twig->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')->setTheme($formView, $theme);
+
+        // BC for Symfony < 3.4 where runtime should be TwigRenderer
+        if (!method_exists(DebugCommand::class, 'getLoaderPaths')) {
+            $twig->getRuntime(TwigRenderer::class)->setTheme($formView, $theme);
+
+            return;
+        }
```
greg0ire: Since when does `FormRenderer` exist?

jordisala1991: Since 3.0 I think, but until 3.4 it was not possible to get it using `getRuntime`. This is a copy paste of the solution applied to AdminBundle, and MediaBundle

greg0ire: http://api.symfony.com/2.3/Symfony/Component/Form/FormRenderer.html

jordisala1991: Yep, it might be since a lot of time, but until 3.4 it was not available to get it from runtime :)

greg0ire: Ok then

PR #305, opened 2017-12-09
Remove tautologies

greg0ire:
I am targeting this branch, because this is pedantic.

This removes a lot of comments that are just tautological. Also, this migrates more classes to Twig namespaces. I ran my script only on `src/Twig` the other day 😅

### DIFF
--- a/src/Form/Type/SimpleFormatterType.php
+++ b/src/Form/Type/SimpleFormatterType.php
@@ -39,20 +39,20 @@ class SimpleFormatterType extends AbstractType
     private $templateManager;
 
     /**
-     * @param ConfigManagerInterface   $configManager   An Ivory CKEditor bundle configuration manager
-     * @param PluginManagerInterface   $pluginManager   An Ivory CKEditor bundle plugin manager
-     * @param TemplateManagerInterface $templateManager An Ivory CKEditor bundle template manager
+     * @param ConfigManagerInterface        $configManager   An Ivory CKEditor bundle configuration manager
+     * @param PluginManagerInterface|null   $pluginManager   An Ivory CKEditor bundle plugin manager
+     * @param TemplateManagerInterface|null $templateManager An Ivory CKEditor bundle template manager
      */
-    public function __construct(ConfigManagerInterface $configManager, PluginManagerInterface $pluginManager = null, TemplateManagerInterface $templateManager = null)
-    {
+    public function __construct(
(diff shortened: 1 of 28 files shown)


src/Formatter/Pool.php
```diff
@@ -14,9 +14,9 @@
 use Psr\Log\LoggerAwareInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
-use Twig_Environment;
-use Twig_Error_Syntax;
-use Twig_Sandbox_SecurityError;
+use Twig\Environment;
+use Twig\Error\SyntaxError as TwigSyntaxError;
```
core23: Why did you use an alias?

greg0ire: Hmmm I don't know why rector did this, let's remove it.

greg0ire:
ping @core23

greg0ire:
ping @core23

src/Form/Type/SimpleFormatterType.php
```diff
@@ -39,20 +38,20 @@ class SimpleFormatterType extends AbstractType
     private $templateManager;
 
     /**
-     * @param ConfigManagerInterface   $configManager   An Ivory CKEditor bundle configuration manager
-     * @param PluginManagerInterface   $pluginManager   An Ivory CKEditor bundle plugin manager
-     * @param TemplateManagerInterface $templateManager An Ivory CKEditor bundle template manager
+     * @param ConfigManagerInterface    $configManager   An Ivory CKEditor bundle configuration manager
+     * @param ?PluginManagerInterface   $pluginManager   An Ivory CKEditor bundle plugin manager
```
core23: This is wrong. Should be:

```
PluginManagerInterface|null   $pluginManager   An Ivory CKEditor bundle plugin manager
```

greg0ire: Are you sure this is wrong? See https://github.com/phpDocumentor/TypeResolver/pull/35/files#diff-58912ee2ce6f273d72422833d8e0721fR48

core23: Never seen this before... What style is used inside symfony code?

greg0ire:
@core23 I reverted to `|null` because apparently people don't do that (yet, I asked), and the codebase isn't 7.1 only yet. Please review again.

PR #313, opened 2018-01-15
use new sonata admin config nodes instead of overwriting template

OskarStark:
simplified docs

### DIFF
--- a/docs/reference/formatter_widget.rst
+++ b/docs/reference/formatter_widget.rst
@@ -11,7 +11,6 @@ Simple Formatter Wiget
 
 The ``sonata_simple_formatter_type_selector`` widget has been implemented to allow developers to force the input formatter value.
 
-
 .. code-block:: php
 
     <?php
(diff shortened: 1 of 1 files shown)


docs/reference/formatter_widget.rst
```diff
@@ -127,81 +126,27 @@ If you stop here, the most interesting part will not be present. Let's edit some
     If you want to see an actual implementation of this can, you can have a look at the ``SonataNewsBundle``.
 
 
-Dynamic Input
```
jordisala1991: Why are you removing this block?

docs/reference/formatter_widget.rst
```diff
@@ -127,81 +126,27 @@ If you stop here, the most interesting part will not be present. Let's edit some
...
-        form:
-            resources:
-                # ...
-                - 'SonataFormatterBundle:Form:formatter.html.twig'
-
-
-Make sure the ``SonataFormatterBundle:Form:formatter.html.twig`` is set.
-This template contains custom javascript code to load the selected text editor.
-
-You also need to include some assets in your template:
```
jordisala1991: Probably this is what you want to remove

docs/reference/formatter_widget.rst
OskarStark: IMO it is duplicate and the same to achieve both

OskarStark: all the stuff is already named in the installation section, so it is duplicate content
https://github.com/sonata-project/SonataFormatterBundle/blob/3.x/docs/reference/installation.rst#configuration

jordisala1991:
Thank you @OskarStark

PR #314, opened 2018-01-15
simplified installation instructions

OskarStark:
We only support SF >= 2.8


docs/reference/formatter_widget.rst
```diff
@@ -6,8 +6,8 @@
 One recurrent need is to provide some kind of rich user interface to edit content.
 However, depending on the final target the content can have different formats: ``markdown``, ``textile``, ``bbcode``, ``rawhtml``, ``richhtml``.
 
-Simple Formatter Wiget
-======================
+Simple Formatter Widget
```
greg0ire: Hahaha good catch! I guess a Wiget must be a small wig!

PR #318, opened 2018-01-27
Update installation documentation for Symfony Flex

kunicmarko20:
I am targeting this branch, because I am updating documentation for Symfony Flex.

## Subject

Updating documentation to adopt new flex structure.

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -8,9 +8,27 @@ To begin, add the dependent bundles:
 
 .. code-block:: bash
 
-    $ php composer.phar require sonata-project/formatter-bundle
+    $ composer require sonata-project/formatter-bundle
 
-Register the bundles in ``app/AppKernel.php``:
+Register the bundles in ``bundles.php`` file:
+
+.. code-block:: php
+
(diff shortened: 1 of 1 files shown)


kunicmarko20:
SonataFormatter depends on  `IvoryCKEditorBundle`  and that bundle does not have support for symfony4 and there are no maintainers (https://twitter.com/egeloen/status/949356450962399235). So we will have a problem with Formatter and any bundle that depends on it, hopefully, someone will take over `IvoryCKEditorBundle`.

The other thing I wanted to ask is, can't we add twig configuration over compiler pass?

```
# Twig Configuration
twig:
    debug:            "%kernel.debug%"
    strict_variables: "%kernel.debug%"

    form_themes:
        - '@SonataFormatter/Form/formatter.html.twig'
```

Something like I did [here](https://github.com/kunicmarko20/ColorPickerBundle/blob/master/DependencyInjection/Compiler/ResourceCompilerPass.php). I can provide a PR if you want to do it this way.

greg0ire:
Thanks @kunicmarko20 !

PR #319, opened 2018-02-03
Declare dependencies explicitly

jordisala1991:

I consider this fixes this other PR https://github.com/sonata-project/dev-kit/pull/385 (because that was related to this bundle)

## Subject

This changes nothing on what gets installed when using this bundle. But declaring explicit dependencies makes you use the versions correctly even when doing prefer-lowest. That makes the build faster, and improves the consistency on this bundle, because we know what is needed for this bundle to work exactly.

Labelled as pedantic cause it will change nothing for normal users

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,9 +28,17 @@
         "knplabs/knp-markdown-bundle": "^1.4",
         "sonata-project/block-bundle": "^3.11",
         "sonata-project/core-bundle": "^3.0",
+        "symfony/config": "^2.8 || ^3.2 || ^4.0",
+        "symfony/dependency-injection": "^2.8 || ^3.2 || ^4.0",
+        "symfony/event-dispatcher": "^2.8 || ^3.2 || ^4.0",
         "symfony/form": "^2.8 || ^3.2 || ^4.0",
         "symfony/framework-bundle": "^2.8 || ^3.2 || ^4.0",
+        "symfony/http-foundation": "^2.8 || ^3.2 || ^4.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -25,12 +25,20 @@
     "require": {
         "php": "^5.6 || ^7.0",
         "egeloen/ckeditor-bundle": "^4.0 || ^5.0",
-        "knplabs/knp-markdown-bundle": "^1.4",
+        "knplabs/knp-markdown-bundle": "^1.7",
         "sonata-project/block-bundle": "^3.11",
-        "sonata-project/core-bundle": "^3.0",
+        "sonata-project/core-bundle": "^3.9",
+        "symfony/config": "^2.8 || ^3.2 || ^4.0",
```
kunicmarko20: You can probably then increase Symfony to `^3.3` also.

composer.json
```diff
@@ -25,12 +25,20 @@
     "require": {
         "php": "^5.6 || ^7.0",
         "egeloen/ckeditor-bundle": "^4.0 || ^5.0",
-        "knplabs/knp-markdown-bundle": "^1.4",
+        "knplabs/knp-markdown-bundle": "^1.7",
```
OskarStark: why this change?

jordisala1991: Because I was trying to reduce the matrix of possible options when doing prefer-lowest to have faster builds

OskarStark: 👍

kunicmarko20:
Thanks @jordisala1991 !

PR #327, opened 2018-04-21
Fix tests

kunicmarko20:
## Subject

Template registry broke some tests.

### DIFF
--- a/tests/Controller/CkeditorAdminControllerTest.php
+++ b/tests/Controller/CkeditorAdminControllerTest.php
@@ -13,6 +13,8 @@ namespace Sonata\FormatterBundle\Tests\Controller;
 
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
+use Sonata\AdminBundle\Templating\TemplateRegistry;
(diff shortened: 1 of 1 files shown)


tests/Controller/CkeditorAdminControllerTest.php
```diff
@@ -132,9 +134,11 @@ private function configureCRUDController()
         $this->request->get('uniqid')->shouldBeCalled();
         $this->container->get('sonata.admin.pool')->willReturn($pool->reveal());
         $this->container->get('sonata.admin.breadcrumbs_builder')->willReturn($breadcrumbsBuilder->reveal());
+        $this->container->get('admin_code.template_registry')->willReturn($templateRegistry->reveal());
```
greg0ire: These tests break because we mock things that we don't own, and this fix adds more things to mock… it would be great to avoid that if we can.

kunicmarko20:
fixed here also.

kunicmarko20:
Oh, we need to bump dependencies or should I add an "if exists"?

greg0ire:
Thanks @kunicmarko20!

PR #331, opened 2018-04-29
Move from Ivory to FOSCkeditor

kunicmarko20:
I am targeting this branch, because this is a bc fix that would allow formatter with sf4

## Changelog

```markdown
### Deprecated
- using of IvoryCKEditorBundle
```

## To do

- [x] Wait for stable release of [FOSCKEditorBundle](https://github.com/FriendsOfSymfony/FOSCKEditorBundle)
- [x] Update tests

## Subject

We took over IvoryCKEditorBundle and now it is under FOS, in the next few days there will be a stable release, this is just to check if tests pass.


kunicmarko20:
DO NOT MERGE.

src/Form/Type/SimpleFormatterType.php
```diff
@@ -44,10 +44,10 @@ class SimpleFormatterType extends AbstractType
...
 
     /**
-     * @param ConfigManagerInterface         $configManager    An Ivory CKEditor bundle configuration manager
-     * @param PluginManagerInterface|null    $pluginManager    An Ivory CKEditor bundle plugin manager
-     * @param TemplateManagerInterface|null  $templateManager  An Ivory CKEditor bundle template manager
-     * @param StylesSetManagerInterface|null $stylesSetManager An Ivory CKEditor bundle styles set manager
+     * @param ConfigManagerInterface         $configManager    An FOS CKEditor bundle configuration manager
+     * @param PluginManagerInterface|null    $pluginManager    An FOS CKEditor bundle plugin manager
+     * @param TemplateManagerInterface|null  $templateManager  An FOS CKEditor bundle template manager
+     * @param StylesSetManagerInterface|null $stylesSetManager An FOS CKEditor bundle styles set manager
```
greg0ire: `s/An/A`

greg0ire: Same above

kunicmarko20:
@core23 Well, we provided an abstraction to Ivory CKEditor, so if anyone was depending on it without explicitly adding it as a requirement, it is not our fault? I discussed this with @greg0ire

kunicmarko20:
And we are green ✅, cc @sonata-project/contributors

kunicmarko20:
hm, you mean if you are extending some of formatter classes and you change something? Ye, I think that would be a BC break. Should we move this to master then and do a major after we clean the master branch?

greg0ire:
Wait what? Which classes? Can you give an example please?

kunicmarko20:
If someone extended some of our services and overwrote it as a services, he would have to typehint agents IvoryCKEditor namespace, their app would brake with this change as our classes are not final and they can do that.

greg0ire:
I see. How about using `class_alias` in Sonata, or maybe even in your bundle?

kunicmarko20:
We decided we didn't want to do that.

So, I rebase on master and we consider doing a major of formatter?

kunicmarko20:
And not sure class alias for 3rd party should be in sonata?

greg0ire:
Yeah you're right, we probably shouldn't define aliases for our other classes… so let's go with a major I guess :'(

kunicmarko20:
Will rebase to master, clean master and make everything final? Does that sound ok?

src/Form/Type/FormatterType.php
```diff
@@ -57,9 +57,9 @@ class FormatterType extends AbstractType
     private $templateManager;
 
     /**
-     * @param ConfigManagerInterface|null   $configManager   An Ivory CKEditor bundle configuration manager
-     * @param PluginManagerInterface|null   $pluginManager   An Ivory CKEditor bundle plugin manager
-     * @param TemplateManagerInterface|null $templateManager An Ivory CKEditor bundle template manager
+     * @param ConfigManagerInterface|null   $configManager   A FOS CKEditor bundle configuration manager
+     * @param PluginManagerInterface|null   $pluginManager   A FOS CKEditor bundle plugin manager
+     * @param TemplateManagerInterface|null $templateManager A FOS CKEditor bundle template manager
```
greg0ire: Please allow FOS or Ivory here

src/DependencyInjection/SonataFormatterExtension.php
```diff
@@ -41,7 +41,7 @@ public function load(array $configs, ContainerBuilder $container)
 
         $bundles = $container->getParameter('kernel.bundles');
 
-        if (isset($bundles['IvoryCKEditorBundle'])) {
+        if (isset($bundles['FOSCKEditorBundle'])) {
```
greg0ire: Please check both bundles, and maybe throw if both are registered.

composer.json
```diff
@@ -24,7 +24,7 @@
     ],
     "require": {
         "php": "^5.6 || ^7.0",
-        "egeloen/ckeditor-bundle": "^4.0 || ^5.0 || ^6.0",
```
greg0ire: Please restore this

composer.json
```diff
@@ -24,7 +24,7 @@
...
+        "friendsofsymfony/ckeditor-bundle": "^1.0",
```
greg0ire: Maybe move this to suggest?

src/DependencyInjection/SonataFormatterExtension.php
greg0ire: Please trigger a deprecation iff Ivory is registered

src/Resources/config/form.xml
```diff
@@ -5,16 +5,16 @@
             <tag name="form.type" alias="sonata_formatter_type"/>
             <argument type="service" id="sonata.formatter.pool"/>
             <argument type="service" id="translator"/>
-            <argument type="service" id="ivory_ck_editor.config_manager"/>
-            <argument type="service" id="ivory_ck_editor.plugin_manager"/>
-            <argument type="service" id="ivory_ck_editor.template_manager"/>
+            <argument type="service" id="fos_ck_editor.config_manager"/>
+            <argument type="service" id="fos_ck_editor.plugin_manager"/>
+            <argument type="service" id="fos_ck_editor.template_manager"/>
```
greg0ire: Please make this dynamic somehow (either do it in php or load an xml file conditionally)

src/SonataFormatterBundle.php
```diff
@@ -30,7 +30,7 @@ public function boot()
     public function registerFormMapping()
     {
         FormHelper::registerFormTypeMapping([
-            'ckeditor' => 'Ivory\CKEditorBundle\Form\Type\CKEditorType',
+            'ckeditor' => 'FOS\CKEditorBundle\Form\Type\CKEditorType',
```
greg0ire: Please make this dynamic

greg0ire:
> Cant we move everything to require-dev and let developer choose between ivory or fos? In a BC way?

This will crash on update I'm afraid.

src/Form/Type/FormatterType.php
kunicmarko20: Remove typehint and add an `if` in method?

src/DependencyInjection/SonataFormatterExtension.php
kunicmarko20: So if they install fos on their own, we should use fos config else we continue doing it like now but throw deprecation

kunicmarko20: and on next major we will replace ivory with fos

src/Resources/config/form.xml
kunicmarko20: Make 2 `form.xml` one for fos, one for ivory?

composer.json
kunicmarko20: If I am going to check types, shouldn't it be in require dev?

kunicmarko20:
Need to update tests and should I revert docs to ivory or should I mention both?

kunicmarko20:
wouldn't this still force IvoryCKEditor because it is still in require part of composer? and it still fails to install on sf4 :disappointed:

greg0ire:
> Need to update tests and should I revert docs to ivory or should I mention both?

Only mention the right way please.

greg0ire:
> wouldn't this still force IvoryCKEditor because it is still in require part of composer? and it still fails to install on sf4

Ah that's right, I forgot that… but there is still hope: How about using the `replaces` key in FOS's composer.json?

kunicmarko20:
TIL 😄 cc @ElectricMaxxx we should add `replace` in composer.json

greg0ire:
Make sure to discuss it and not rush though, I'm not sure of the implications, since you changed the namespaces, it's not the same functionality, technically… but anyone using your package is aware they need to change the namespaces I suppose.

ElectricMaxxx:
I currently do not understand the issue. As you do not expose a third party Interface without wrapping it, the user would not have any change, or am i right. When somebody relies on an Ivory class requiring your bundle only, the it is no BC from my POV.

kunicmarko20:
@ElectricMaxxx we have typehints in constructors of our form types, if someone is extending and overriding that form types, his app would break if we change the type of class in extended class. So, I will check out that replace key later more or we can do a Major release maybe.

src/DependencyInjection/SonataFormatterExtension.php
```diff
@@ -41,8 +41,10 @@ public function load(array $configs, ContainerBuilder $container)
 
         $bundles = $container->getParameter('kernel.bundles');
 
-        if (isset($bundles['IvoryCKEditorBundle'])) {
-            $loader->load('form.xml');
+        if (isset($bundles['FOSCKEditorBundle'])) {
+            $loader->load('form_fos.xml');
+        } elseif (isset($bundles['IvoryCKEditorBundle'])) {
+            $loader->load('form_ivory.xml');
```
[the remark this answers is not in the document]
kunicmarko20: then I just remove stuff from ctor and throw 1 deprecation here?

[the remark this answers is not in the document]
greg0ire: yup. People don't need to be told to move to FOS 23409823 times :P

greg0ire:
> What about removing the type hint an do a manual instanceof, doing so you could easily deprecate that Argument type. Removing a type would be no BC break.

Uuuuh… that's exactly what I suggested and has been done

kunicmarko20:
Seems that we are going to revert replace https://github.com/FriendsOfSymfony/FOSCKEditorBundle/pull/97 after all, so, should I rebase this on master? cc @sonata-project/contributors

kunicmarko20:
Closing this as there is no way to implement this without BC Break, will just change it on master.

greg0ire:
Whaaaat? Why? If you don't use `replace` , you can still get this, even if it means having both bundles install. Don't assume next major will happen soon, I've made that kind of assumptions in the pass, it's usually wrong.

docs/reference/installation.rst
```diff
@@ -4,7 +4,20 @@
...
+
+.. code-block:: json
+
+    "replace": {
+        "egeloen/ckeditor-bundle": "*"
+    }
+
+And after that you need to install `FOSCKEditorBundle`_.
+
+Install SonataFormatterBundle:
```
kunicmarko20: This part is the only new thing.

kunicmarko20:
Thanks to @greg0ire we found a way to keep this on stable and avoid BC Break.

greg0ire:
The docs job failed

kunicmarko20:
should be ok now

greg0ire:
@core23 please unblock this

greg0ire:
Thanks @kunicmarko20 !

kunicmarko20:
It would be a bc break 😞

PR #332, opened 2018-05-03
Use renderWithExtraParams() instead of render()

greg0ire:

## Changelog

```markdown
### Fixed
- deprecations from the admin bundle about `render` vs `renderWithExtraParams`
```
## Subject

render() has been planned for removal and should no longer be used.


jordisala1991:
Thank you @greg0ire

PR #341, opened 2018-05-16
Fix ckeditor toolbar config

hug963:
I am targeting this branch, because this is a bug fix and BC.

## Changelog

```markdown
### Fixed
- Ckeditor toolbar config not loading
```

## Subject

v5 of https://github.com/egeloen/IvoryCKEditorBundle introduced a [toolbar manager](https://github.com/egeloen/IvoryCKEditorBundle/commit/7ef1c96f6958b458545f336a10178f033764fd5f) to load the toolbar configuration.

This was never implemented in this bundle although it is compatible with V5 and V6 of the IvoryCKEditorBundle.


hug963:
In V4 it was not broken at the toolbar was resolved at compile time.

It only stopped working in v5 when the toolbar manager was implemented.

I added `on-invalid="null"` in the config and tested locally with V4 installed and it works as it should!

greg0ire:
Thanks @hug963 !

PR #343, opened 2018-05-19
Manual Merge

kunicmarko20:
Just approve

### DIFF
--- a/src/Form/Type/FormatterType.php
+++ b/src/Form/Type/FormatterType.php
@@ -16,6 +16,7 @@ namespace Sonata\FormatterBundle\Form\Type;
 use FOS\CKEditorBundle\Model\ConfigManagerInterface;
 use FOS\CKEditorBundle\Model\PluginManagerInterface;
 use FOS\CKEditorBundle\Model\TemplateManagerInterface;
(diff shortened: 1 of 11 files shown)


src/Form/Type/FormatterType.php
```diff
@@ -58,18 +59,25 @@
...
+     */
+    private $toolbarManager;
+
     public function __construct(
         Pool $pool,
         TranslatorInterface $translator,
         ConfigManagerInterface $configManager,
         ?PluginManagerInterface $pluginManager = null,
-        ?TemplateManagerInterface $templateManager = null
+        ?TemplateManagerInterface $templateManager = null,
```
kunicmarko20: are we sure this should stay nullable on master?

greg0ire:
Argh, FlintCI disapproves

src/Form/Type/FormatterType.php
greg0ire: Maybe not, but this is not the PR to change it, don't do [evil merges](https://blog.joolsadams.org/2013/02/15/evil-merges-in-git/)

kunicmarko20: I know I know, I just now saw it and was wondering :smile:

kunicmarko20:
Fixed.

PR #336, opened 2018-05-06
Add type hints

greg0ire:
I am targeting this branch, because this is a BC break.

## Changelog

```markdown
### Added
- PHP 7.1 type hints
```

PR #347, opened 2018-06-03
Remove some typehints

greg0ire:
In 155337dbece5cbaefac716da7e3b3b93096f7007, I made the wrong assumption
that Twig would have type hints, and introduced a BC break a class that
implements a Twig interface. This removes those type hints, and adds a
test to be sure those cannot be added again by mistake.

I am targeting this branch, because this fix is only relevant on master.

Refs #336 

## Changelog

None, since nothing is released yet.

### DIFF
--- a/src/Twig/SecurityPolicyContainerAware.php
+++ b/src/Twig/SecurityPolicyContainerAware.php
@@ -71,7 +71,7 @@ final class SecurityPolicyContainerAware implements SecurityPolicyInterface
         $this->extensions = $extensions;
     }
 
-    public function checkSecurity(array $tags, array $filters, array $functions): void
+    public function checkSecurity($tags, $filters, $functions): void
     {
         $this->buildAllowed();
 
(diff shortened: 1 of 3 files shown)


src/Twig/SecurityPolicyContainerAware.php
```diff
@@ -71,7 +71,7 @@ public function __construct(ContainerInterface $container, array $extensions = [
         $this->extensions = $extensions;
     }
 
-    public function checkSecurity(array $tags, array $filters, array $functions): void
+    public function checkSecurity($tags, $filters, $functions): void
```
jmontoyaa: The return type declarations "bool", "void", are not present in the interface:

https://github.com/twigphp/Twig/blob/v2.4.8/lib/Twig/Sandbox/SecurityPolicyInterface.php

greg0ire: @jmontoyaa yes, and?

jmontoyaa: my bad, it works well :+1:

PR #368, opened 2018-10-03
Added null check to FormatterListener

core23:
I am targeting this branch, because this is a patch.

## Changelog

```markdown
### Fixed
- Catch null values in FormatterListener
```

## Subject

The `FormatterListener` was failing if you tried to transform a null value.


greg0ire:
Thanks @core23 !

PR #389, opened 2018-12-20
Removed all specific formatters

### LINKED ISSUE #365: [RFC] Remove all (optional) dependencies
At the moment, this bundle provides just one new form component (FormatterType) which supports many formatters (richtext, twig, ckeditor, markdown). Also you don't need everything, the dependencies would be loaded. This could cause some problems (see CKEditor).

My idea is just to remove every formatter, except some dummy input=output formatter, and move everything specific to seperate bundles (ckeditor-formatter-bundle, markdown-formatter-bundle, ...). This could be a great extensions point for many different formatter types.

When moving all stuff away from this bundle, we could also think about moving the remaining parts to the form-extensions repo.

WDYT @sonata-project/contributors

core23:
## Subject

This PR will remove all formatters from this repo to support more formatters or change them without doing a major release in the future.

I am targeting this branch, because this is a huge BC break.

Closes #365

## Changelog

```markdown
### Removed
- Removed all specific formatters
```

After merging this PR, we could create some new bundles with specific formatters:

- [ ] Create SonataCkeditorFormatterBundle
- [ ] Create SonataMarkdownFormatterBundle 
- [ ] Create SonataTwigFormatterBundle


kunicmarko20:
So what is the idea behind this? Have like an interface and adapters per type of the editor? Maybe adapter repositories with the same idea? Do we maybe delegate everything to the end user?

kunicmarko20:
I just saw the issue, not sure how I didn't see it before. I am all for this.

OskarStark:
hmm and how do I know how to add ckeditor and use it with the media bundle?

core23:
> hmm and how do I know how to add ckeditor and use it with the media bundle?

We have to create a new repo (e.g. SonataCkeditorAdapter) for this. 

This might look like a repo overkill, but it allows us some more flexible maintenance for each project without doing huge BC breaks every time. We will be more flexible with version management.

greg0ire:
Looks OK, what about your todo list?

core23:
> Looks OK, what about your todo list?

"This can be done after merging this"

We don't need this for the first start. Also we don't know if we still need support for all formatters.

greg0ire:
> "This can be done after merging this"

"This" is ambiguous in both cases :sweat_smile:

core23:
Updated the description 🤣 

After merging this one, we could create a new stable release and could start creating new bundles. For a short time, we should support two stable version (with and without default formatters).

greg0ire (approved):

OskarStark (approved):

core23:
> > After merging this one, we could create a new stable release and could start creating new bundles.
> 
> You are talking about a major? and we supporting 4.x and 5.x?

Yes, unless we have all new formatter bundles

kunicmarko20:
I guess this is fine, we probably can't do a new major until we have at least one new adapter lib?

kunicmarko20 (approved):

OskarStark (approved):

core23:
> I guess this is fine, we probably can't do a new major until we have at least one new adapter lib?

No, we should create a major, otherwise we can't create an adapter. There is also one default `TextFormatter` inside this bundle.

jordisala1991:
What should we do about this PR?

I have 2 main concerns about it:
- The Pr talks about formatters but it also removed a lot of code to connect with media bundle without any advise or replacement provided.
- It mentions the creation of new bundles with less code. Even if that is a good idea, we dont have any of that created yet and we are trying to reduce the number of bundles to maintain.

IMO we should:
- Revert the code related to media bundle, I dont know why it was removed.
- Decide which formatters are important to us and restore support here directly with dev dependencies, the other formatters will just be removed and gone (people outside sonata can decide to reimplement them if they want). I would consider only restoring the ckeditor formatter.

Wdyt @sonata-project/contributors ?

VincentLanglet:
> Wdyt @sonata-project/contributors ?

I agree.
We should also check if there are similar PR.
If we forgot about why we did some things in the master branch, we might consider reseting it with the stable one and starting again...

core23:
> If we forgot about why we did some things in the master branch, we might consider reseting it with the stable one and starting again...

The idea was to remove all third party components from this project, because they will restrict us from maintaining this project. A huge problem across this project (and the admin project) is that it relies on many outdated assets (e.g. jQuery, CKEditor, ...).

Because there was no good asset management strategy, I wanted to move all of this stuff to small atomic projects. This way we do not need to release a new major versions of the formatter bundle as soon as we update the CKEditor or some Markdown version.

The basic idea is similar to the `psr/log` library. This library provides an interface that is stable and won't change quite often. If you need some specific logging / formatter you can load a third party project that provides an HTML or Markdown editor.

jordisala1991:
> > If we forgot about why we did some things in the master branch, we might consider reseting it with the stable one and starting again...
> 
> 
> 
> The idea was to remove all third party components from this project, because they will restrict us from maintaining this project. A huge problem across this project (and the admin project) is that it relies on many outdated assets (e.g. jQuery, CKEditor, ...).
> 
> 
> 
> Because there was no good asset management strategy, I wanted to move all of this stuff to small atomic projects. This way we do not need to release a new major versions of the formatter bundle as soon as we update the CKEditor or some Markdown version.
> 
> 
> 
> The basic idea is similar to the `psr/log` library. This library provides an interface that is stable and won't change quite often. If you need some specific logging / formatter you can load a third party project that provides an HTML or Markdown editor.

Yes, but again, who is going to create / maintain those other packages? Releasing a new version of this without the other work done is not possible, because as you said this library is now some interfaces and a few glue code.

To me it does not make sense to leave this completely empy. I do agree on removing some of the formatters and leave only raw and ckeditor, since those should probably the most used. And leave the rest for outside sonata.

VincentLanglet:
Anyway if we want to remove something to move it somewhere else, we should add it somewhere else before remove it here.

@jordisala1991 you talk about code related to media bundle.
Couldn't this code be added to media bundle ? And the formatter one would be a dev-dependency

lukepass:
I agree with Vincent that the CKEditor integration has been removed too quickly without a replacement. At the moment there is no working bundle to integrate CKEditorBundle with SonataMediaBundle (CoopTilleulsCKEditorSonataMediaBundle is abandoned).
I think that it should at least maintain the file upload routes so that we can integrate the editor in another way.

Thanks.

EDIT: I am referring to this part of the SonataFormatterBundle documentation: https://docs.sonata-project.org/projects/SonataFormatterBundle/en/4.x/reference/ckeditor_medias/

VincentLanglet:
Since the strategy was to provide those formatter in another bundle, with the todo list
- Create SonataCkeditorFormatterBundle
- Create SonataMarkdownFormatterBundle
- Create SonataTwigFormatterBundle

But we avoid creating new bundle now, so there is no solution for those formatter now.
I think we can re-add all of them until a better/another solution is found.
If one day we want to create new bundle, it should be done before removing them from here.

@lukepass Do you want to open the PR reverting this one ?

jordisala1991:
We either revert all of them at once or we do it when someone ask for them. 
Maybe no one is using the Markdown or Twig thing.

lukepass:
I've only ever used the CKEditor formatter because it's the best way to integrate it with SonataMedia. What should I do to revert it? Look at the commits, find the relevant code and make a new commit with the old code? Thanks.

haivala:
I'm taking a step back and looking at this bundle.

I would imagine that the most common use case for the formatter is the CkEditor one as it gives the user WYSIWYG editor and that helps user to understand the formatting in the actual place where the content is going to be used (if not in place)

That said the purpose of this bundle is just that: give non technical user ability to edit content and insert media in middle of text.

Is there even better way to do it?

Does this help to find solution to the issue?

VincentLanglet:
> What should I do to revert it? Look at the commits, find the relevant code and make a new commit with the old code? Thanks.

Yes, re-adding all the relevant code is fine.

core23:
> Yes, re-adding all the relevant code is fine.

:+1: 

Maybe you can use symfony-ux / stimulus to create the widgets instead of inlining the CKEditor JavaScript to every page.

VincentLanglet:
> I've only ever used the CKEditor formatter because it's the best way to integrate it with SonataMedia. What should I do to revert it? Look at the commits, find the relevant code and make a new commit with the old code? Thanks.

Cf https://github.com/sonata-project/SonataFormatterBundle/pull/646

We gonna only add Ckeditor support back @lukepass, PR is welcomed

VincentLanglet:
Let's continue the discussion here: https://github.com/sonata-project/SonataFormatterBundle/issues/647

PR #417, opened 2019-08-18
Add new interface for extendable formatters

core23:
## Subject

Some formatters do not support extensions (e.g. https://github.com/sonata-project/SonataFormatterBundle/blob/4.x/src/Formatter/TwigFormatter.php#L56), so we should split the existing interface

## Changelog

```markdown
### Added
- Added new `Formatter` interface
- Added new `ExtendableFormatter` interface
```

### DIFF
--- a/src/Formatter/Formatter.php
+++ b/src/Formatter/Formatter.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\FormatterBundle\Formatter;
+
+interface Formatter
+{
+    /**
+     * @return string
+     */
+    public function transform(string $text);
+}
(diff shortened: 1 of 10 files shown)


src/Formatter/FormatterInterface.php
```diff
@@ -13,16 +13,15 @@
 
 namespace Sonata\FormatterBundle\Formatter;
 
-use Sonata\FormatterBundle\Extension\ExtensionInterface;
+@trigger_error(
+    'The '.__METHOD__.' interface is deprecated since sonata-project/formatter-bundle 4.x, to be removed in 5.0. '.
```
phansys:
```suggestion
    'The '.__NAMESPACE__.' interface is deprecated since sonata-project/formatter-bundle 4.x, to be removed in 5.0. '.
```

src/Formatter/Formatter.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\FormatterBundle\Formatter;
+
+interface Formatter
+{
+    public function transform(string $text);
```
phansys: Maybe we could tell that with the next major, `void` will be declared as return type.

core23:
Ping @OskarStark

PR #434, opened 2020-02-06
Manual merge

greg0ire:


greg0ire:
PHPStan issues also affect the stable branch I think, and should be addressed there IMO

greg0ire:
Actually, one of them is because we no longer require `admin-bundle` on master.

core23:
> Actually, one of them is because we no longer require `admin-bundle` on master since #389
> @core23 please advise 🙏
> 
> If fixed the other one in #435

We should add the admin bundle as a dev dependency. For the next major release, we should use the block bundle FormMapper

greg0ire:
@core23 done, it works :)

PR #452, opened 2020-04-14
DevKit updates for 4.x branch

SonataCI:


greg0ire:
The PHPStan job is failing because `ext-zip` misses on the docker image we are using. Related issue: https://github.com/OskarStark/phpstan-ga/issues/9

greg0ire:
PHPStan works, and on top of that, it found a bug, which I just fixed :)

PR #449, opened 2020-02-25
Update FOSCKEditor dependency to 2.0

psyray:
## Subject

This PR update dependency to FOSCKEditor v2.0 to remove a lot of deprecation warnings in SF 4 and prepare this bundle to SF 5

I am targeting this branch, because it's the main branch

Closes #447 

## Changelog

```markdown
### Changed
Change form types to use the main fos_ck_editor.configuration service

### Removed
Deprecated FOS CKEditor services

### Fixed
Some SF deprecation warnings (~50)

```

## To do
    
- [ ] Update the tests
- [ ] More functional tests
- [ ] Add an upgrade note

In my project, CKEditor is loading fine, changing Toolbar works well.
Assets need to be refreshed.
I need more testing & feedback

PR #467, opened 2020-06-30
Update FOSCKEditor dependency to 2.0

### LINKED ISSUE #447: Upgrade FOS CKEditor dependency to 2.0
## Feature Request

Is it possible to update the **friendsofsymfony/ckeditor-bundle** to 2.0 ?
To fix some deprecation in Symfony and prepare bundle to 5.0 compatibility

psyray:
## Subject

This PR update dependency to FOSCKEditor v2.0 to remove a lot of deprecation warnings in SF 4 and prepare this bundle to SF 5

I am targeting this branch, because it's the main branch

Replace #449 
Closes #447 

## Changelog

```markdown
### Changed
Change form types to use the main fos_ck_editor.configuration service

### Removed
Deprecated FOS CKEditor services

### Fixed
Some SF deprecation warnings (~50)

```

## To do
    
- [x] Update the tests

In my project, CKEditor is loading fine, changing Toolbar works well.
No special upgrade for SonataFormatterBundle, only assets need to be refreshed.
```bash
bin/console assets:install
```
Check upgrade note in FOSCKEditorBundle repository, only one line should be added in the twig config file to render the editor
https://github.com/FriendsOfSymfony/FOSCKEditorBundle/blob/2.x/UPGRADE-2.0.md

### DIFF
--- a/src/Form/Type/FormatterType.php
+++ b/src/Form/Type/FormatterType.php
@@ -13,10 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\FormatterBundle\Form\Type;
 
-use FOS\CKEditorBundle\Model\ConfigManagerInterface;
-use FOS\CKEditorBundle\Model\PluginManagerInterface;
-use FOS\CKEditorBundle\Model\TemplateManagerInterface;
-use FOS\CKEditorBundle\Model\ToolbarManagerInterface;
+use FOS\CKEditorBundle\Config\CKEditorConfigurationInterface;
 use Sonata\FormatterBundle\Form\EventListener\FormatterListener;
 use Sonata\FormatterBundle\Formatter\PoolInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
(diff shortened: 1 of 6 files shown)


src/Form/Type/FormatterType.php
```diff
@@ -44,39 +41,18 @@ final class FormatterType extends AbstractType
...
-    /**
-     * @var ToolbarManagerInterface
-     */
-    private $toolbarManager;
+    private $ckEditorConfiguration;
 
     public function __construct(
         PoolInterface $pool,
         TranslatorInterface $translator,
-        ConfigManagerInterface $configManager,
```
core23, line 69: Not sure if we should handle this as a BC break. The class is final, but you could create a new instance with the *old* signature

psyray: @core23 
Don't really understand, this question is for me or the reviewers ?

core23: It's a question for the other @sonata-project/contributors

VincentLanglet: To me it's a BC.

Everytime we had sometimes like this, we tried to support both way to instantiate the class, with a deprecation for the old way.

bmaziere: any news on this @psyray? need some help to get into what's @VincentLanglet explained?

VincentLanglet: @bmaziere Without news, you can try to finish the PR if you're interested ;)

bmaziere: @VincentLanglet sure I can ;) I guess that as old classes/interfaces has been removed in 2.x of friendsofsymfony/ckeditor-bundle and if we want to keep the ability to instanciate the class the old way, we need to allow both version in composer.json right?

VincentLanglet: Reconsidering the situation, I'm not sure we can call this a BC-break, since the we dropped support for CKEditor 1.0. To create new instance with the old signature, you'll need CKEditor 1.x isn't it ? So you'll not be able to install this version.
I'm not surprised by the fact that changes are expected when if the user update his CKEditor version.

WDYT @sonata-project/contributors ? Especially @core23 ?

Is it mandatory to keep supporting the v1 version and make the constructor compatible with both version ?

jordisala1991: IMHO it would be nice to upgrade first one bundle and then another. Otherwise to update to this version you need to update the FOSCKEditorBundle too. 

In this case the compatibility a maybe a bit hard, but it is doable (not like other updates we had to do previously)

VincentLanglet: Ok, so @bmaziere, it would be nice to support `^1.0 || ^2.0`. 
And we would like to keep the ability to instanciate the class the old way and the new way.

You can see how we did (for another bundle) here: https://github.com/sonata-project/SonataAdminBundle/pull/6220/files

psyray: @bmaziere
Thanks to finish my PR, It's holiday time for me :)

jordisala1991:
This can be closed, the 2.0 is already allowed

PR #478, opened 2020-08-20
Manually merge 4.x into master

greg0ire:


composer.json
```diff
@@ -22,7 +22,7 @@
     ],
     "require": {
         "php": "^7.2",
-        "friendsofsymfony/ckeditor-bundle": "^1.0",
+        "friendsofsymfony/ckeditor-bundle": "^1.0 || ^2.0",
```
greg0ire, line 25: :thinking: when I merged, it looked like we just removed the bundle, but then there still is this dependency… why?

greg0ire: FYI it was removed in #389 @bmaziere

greg0ire: cc @core23  I think I will just remove the dependency.

jordisala1991 (approved):
Guess the merge is okay... but having such features removed without a path of migration will be hard to deal with

PR #494, opened 2020-09-18
Use ProphecyTrait to fix warnings

franmomu:
The PHP 8 test was failing because it throws warnings about Prophecy integration being deprecated in PHPunit 9.
I added `phpspec/prophecy-phpunit` to use `ProphecyTrait`. I guess the same should be done in SonataAdminBundle.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,7 @@
     },
     "require-dev": {
         "matthiasnoback/symfony-dependency-injection-test": "^4.1",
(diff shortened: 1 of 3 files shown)


phansys (approved):

franmomu:
Sadly `phpspec/prophecy-phpunit` requires `"php": "^7.3 || ^8",`

PR #520, opened 2020-11-02
Replace prophecy by phpunit mocks

jordisala1991:


tests/Controller/CkeditorAdminControllerTest.php
```diff
@@ -55,148 +55,152 @@ class CkeditorAdminControllerTest extends TestCase
 
     protected function setUp(): void
     {
-        $this->container = $this->prophesize(ContainerInterface::class);
-        $this->admin = $this->prophesize(BaseMediaAdmin::class);
-        $this->request = $this->prophesize(Request::class);
+        $this->container = new Container();
+        $this->admin = $this->createMock(BaseMediaAdmin::class);
+        $this->request = $this->createMock(Request::class);
```
greg0ire, line 60: It would be great to use real objects just like you did on line 58 with the container. Disregard this comment if this is too much work.

greg0ire (approved):

tests/Controller/CkeditorAdminControllerTest.php
jordisala1991: I did on container because it was easier to unmock it rather than try to replace it with createMock. In this case, this mocks requires a bit of rework of the test to be accurate when creating real objects. I would prefer to not have to change it for now.

VincentLanglet (approved):

VincentLanglet:
Thanks @jordisala1991

Issue #643, opened 2022-01-12
Symfony 5 error with "sonata.formatter.text.raw"


lukepass:
### Environment

Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.3 LTS
Release:	20.04
Codename:	focal

#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
sonata-project/admin-bundle              4.5.0           4.7.0           The missing Symfony Admin Gener...
sonata-project/block-bundle              4.9.0           4.9.1           Symfony SonataBlockBundle
sonata-project/cache                     2.2.0           2.2.0           Cache library
sonata-project/doctrine-extensions       1.16.0          1.16.0          Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 4.2.1           4.2.1           Integrate Doctrine ORM into the...
sonata-project/exporter                  2.9.1           2.10.1          Lightweight Exporter library
sonata-project/form-extensions           1.12.4          1.12.4          Symfony form extensions
sonata-project/formatter-bundle          5.x-dev 25ccae3 5.x-dev 25ccae3 Symfony SonataFormatterBundle
sonata-project/media-bundle              4.0.0-RC2       4.0.0-RC2       Symfony SonataMediaBundle
sonata-project/twig-extensions           1.9.1           1.9.1           Sonata twig extensions
```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'symfony/*'
symfony/amqp-messenger             v5.4.0  v5.4.0  Symfony AMQP extension Messenger Bridge
symfony/apache-pack                v1.0.1  v1.0.1  A pack for Apache support in Symfony
symfony/asset                      v5.4.0  v6.0.1  Manages URL generation and versioning of web assets s...
symfony/browser-kit                v5.4.0  v6.0.1  Simulates the behavior of a web browser, allowing you...
symfony/cache                      v5.4.0  v6.0.2  Provides an extended PSR-6, PSR-16 (and tags) impleme...
symfony/cache-contracts            v2.5.0  v3.0.0  Generic abstractions related to caching
symfony/config                     v5.4.0  v6.0.2  Helps you find, load, combine, autofill and validate ...
symfony/console                    v5.4.1  v6.0.2  Eases the creation of beautiful and testable command ...
[...]

PR #645, opened 2022-01-18
Fix class name references in SonataFormatterExtension

jorrit:
## Subject

I think that having `::class` references in calls to `new Definition()` is the right way of creating definitions.

I am targeting this branch, because it's part of moving to Symfony 5.

See #643

### DIFF
--- a/src/DependencyInjection/SonataFormatterExtension.php
+++ b/src/DependencyInjection/SonataFormatterExtension.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace Sonata\FormatterBundle\DependencyInjection;
 
 use Sonata\FormatterBundle\Formatter\ExtendableFormatter;
+use Sonata\FormatterBundle\Twig\Loader\LoaderSelector;
+use Sonata\FormatterBundle\Twig\SecurityPolicyContainerAware;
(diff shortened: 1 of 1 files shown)


jorrit:
I do not understand this code in SonataFormatterExtension.php:

```php
        $env = new Definition(Environment::class, [$loaderSelector, [
            'debug' => false,
            'strict_variables' => false,
            'charset' => 'UTF-8',
        ]]);
....

        if (is_a($env->getClass(), ExtendableFormatter::class)) {

            foreach ($extensions as $extension) {
                $env->addMethodCall('addExtension', [new Reference($extension)]);
            }
        }
```

Isn't `$env->getClass()` always equal to `Environment::class` at this point?

Shouldn't the comparison be like:

`if (is_a($configuration['service'], ExtendableFormatter::class)) {`

VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thanks

jorrit:
Thanks @jordisala1991 . Could you take a look at my comment at https://github.com/sonata-project/SonataFormatterBundle/pull/645#issuecomment-1015256814 ?

PR #648, opened 2022-01-29
Add psalm

VincentLanglet:
pedantic

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -37,3 +37,27 @@ jobs:
 
             - name: PHPStan
               run: vendor/bin/phpstan --memory-limit=1G analyse --error-format=github
+    psalm:
+        name: Psalm
+
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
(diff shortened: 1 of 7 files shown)


VincentLanglet:
@jordisala1991 The psalm error is kinda weird, does it means we don't really support php 8.1 ?

jordisala1991:
I think we do but with sone warnings. But those are related to SonataAdminBundle 3.x right?

VincentLanglet:
Yes we have warning with 8.1 on sonata admin 3 and it's giving a fatal error on psalm here, but not on sonata admin...

jordisala1991:
Sonata Admin 3 is not being tested against php 8.1

VincentLanglet:
> Sonata Admin 3 is not being tested against php 8.1

Oh does that mean that SonataAdmin 3 doesn't support 8.1 and we shouldn't say we support 8.1 on SonataFormatter ?

jordisala1991:
Probably yes, we should remove it from devkit

jordisala1991 (approved):

PR #658, opened 2022-03-13
Increase PHPStan, Psalm, cleanup deps

jordisala1991:
## Subject

I am targeting this branch, because this is for next release.


phpstan-baseline.neon
```diff
@@ -0,0 +1,17 @@
...
+            count: 1
+            path: src/Extension/BaseProxyExtension.php
+
+        -
+            message: "#^Class Twig\\\\Extension\\\\InitRuntimeInterface not found\\.$#"
+            count: 1
+            path: src/Extension/BaseProxyExtension.php
+
+        -
+            message: "#^Access to an undefined property Symfony\\\\Component\\\\Validator\\\\Constraint\\:\\:\\$message\\.$#"
```
VincentLanglet: You need to change

```
public function validate($value, Constraint $constraint): void
    {
        if (!$this->pool->has($value)) {
            $this->context->addViolation($constraint->message);
        }
    }
```
to
```
public function validate($value, Constraint $constraint): void
    {
       if (!$constraint instanceof FormatterConstraint) {
            throw new UnexpectedTypeException($constraint, FormatterConstraint::class);
      }
      if (!$value instanceof ...) {
           throw new UnexpectedValueException($value, ...);
     }


        if (!$this->pool->has($value)) {
            $this->context->addViolation($constraint->message);
        }
    }
```

phpstan-baseline.neon
```diff
@@ -0,0 +1,17 @@
+parameters:
+    ignoreErrors:
+        -
+            message: "#^Call to method initRuntime\\(\\) on an unknown class Twig\\\\Extension\\\\InitRuntimeInterface\\.$#"
+            count: 1
+            path: src/Extension/BaseProxyExtension.php
+
+        -
+            message: "#^Class Twig\\\\Extension\\\\InitRuntimeInterface not found\\.$#"
```
VincentLanglet: Why is it not found ? We should add a comment

src/Extension/BaseProxyExtension.php
```diff
@@ -46,9 +48,16 @@ public function getAllowedMethods(): array
         return [];
     }
 
+    /**
+     * @psalm-suppress UndefinedClass, UndefinedInterfaceMethod
```
VincentLanglet, line 54: If we add the psalm comment here, maybe it's better to add phpstan-ignore here too.

src/Twig/SecurityPolicyContainerAware.php
```diff
@@ -140,16 +140,16 @@ public function checkMethodAllowed($obj, $method): void
 
     /**
      * @param object $obj
-     * @param string $property
```
VincentLanglet, line 143: I opened https://github.com/twigphp/Twig/pull/3664/files ^^

phpstan-baseline.neon
jordisala1991: Fixed.

src/Extension/BaseProxyExtension.php
jordisala1991: Changed

phpstan-baseline.neon
jordisala1991: We should probably deprecate this on 4.x to remove it on 5.0, the interface is deprecated on 2.x of twig, removed on 3.0. I have added a comment in the method.

src/Twig/SecurityPolicyContainerAware.php
jordisala1991: 👍 Since it requires either changing the property name or adding a suppress we will need to come back here once it is merged an released, so I prefer leave it as is and fix it again later.

jordisala1991:
I am not sure on how to fix the last error on psalm:

```
ERROR: DocblockTypeContradiction - src/DependencyInjection/SonataFormatterExtension.php:135:13 - Operand of type false is always false (see https://psalm.dev/155)
        if (is_a($env->getClass(), ExtendableFormatter::class)) {
```

VincentLanglet:
> I am not sure on how to fix the last error on psalm:
> 
> ```
> ERROR: DocblockTypeContradiction - src/DependencyInjection/SonataFormatterExtension.php:135:13 - Operand of type false is always false (see https://psalm.dev/155)
>         if (is_a($env->getClass(), ExtendableFormatter::class)) {
> ```

$env->getClass() is a string no ? To use is_a with a string, I think you have to pass true as third parameter.

src/Twig/Loader/LoaderSelector.php
```diff
@@ -34,35 +34,34 @@ public function __construct(LoaderInterface $stringLoader, LoaderInterface $file
         $this->fileLoader = $fileLoader;
     }
 
-    public function getSource(string $name): string
-    {
-        $source = $this->getLoader($name)->getSource($name);
```
jordisala1991, line 39: This method was deprecated on 1.x and removed on 2.0... Not sure what to do with it, and maybe is relevant because of the following lines. (Maybe that's why some things didn't work in sandbox.

VincentLanglet: I would say it's ok to remove it since it's not in the interface anymore.

VincentLanglet (dismissed):

src/Twig/Loader/LoaderSelector.php
jordisala1991: Take a look at my new commit, not sure if this way would be better, If there are no test about how all this should work, it is harded to know if it is broken.

I see two options:
* We either don't remove this method (yet).
* We move this logic to `getSourceContext`.

On both options it would be nice to start introducing some sort of test to see if any of this works.

VincentLanglet: > We either don't remove this method (yet).

What I don't understand, is why not keeping this method would solve any bugs. To me, if the `getSource` method is not in the interface anymore, it's never called by Twig.

> We move this logic to getSourceContext.

The getSource method was introduced 11 years ago. And the getSourceContext 5 years ago https://github.com/sonata-project/SonataFormatterBundle/commit/c045b76ea532372eb55c3c68cac369e5d5c5a513

I don't know when is used the LoaderSelector, but according to https://github.com/sonata-project/SonataFormatterBundle/blob/1684ad4a2459e6e3ba29721fa954bb29a4d4de15/src/DependencyInjection/SonataFormatterExtension.php#L156-L160 it seems like your new fix might be useful.

My only concern would be "why nobody opened an issue in 5 years".

VincentLanglet:
Can we merge this @jordisala1991 ?

src/Twig/Loader/LoaderSelector.php
jordisala1991: lets keep it this way then?

jordisala1991:
Yep, it looks ready to me. Should I add changelog?

VincentLanglet (approved):

VincentLanglet:
> Yep, it looks ready to me. Should I add changelog?

It seems ok to me like this.

PR #659, opened 2022-03-17
Bump phpstan to level 8 with strict rules.

VincentLanglet:
Pedantic.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     ],
     "homepage": "https://docs.sonata-project.org/projects/SonataFormatterBundle",
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.4 || ^8.0",
         "friendsofsymfony/ckeditor-bundle": "^2.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
         "sonata-project/block-bundle": "^4.0",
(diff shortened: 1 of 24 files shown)


jordisala1991 (dismissed):

VincentLanglet:
We now need to
- Add back the support of Ckeditor
- Use php instead of xml for the config

and we'll be able to create a pre-release.

VincentLanglet:
Also, like reported by @jorrit here https://github.com/sonata-project/SonataFormatterBundle/pull/645#issuecomment-1015256814

I don't understand this code
```
if (null !== $env->getClass() && is_a($env->getClass(), ExtendableFormatter::class, true)) {

            foreach ($extensions as $extension) {
                $env->addMethodCall('addExtension', [new Reference($extension)]);
            }
        }
```

We're checking if `is_a($env->getClass(), ExtendableFormatter::class, true)` and then call `addExtension` on the env.
But the TwigEnvironnement never extends ExtendableFormatter, and addExtension doesn't exists on TwiEnvironnement.
Shouldn't the check instead be done on 
```
$container->getDefinition($$formatterConfig['service'])
```

WDYt @jordisala1991 ?

cc @core23 you modified this code https://github.com/sonata-project/SonataFormatterBundle/commit/f4497f0ab577949c9d23b57f988925bfa9ffc4a0

PR #679, opened 2022-06-12
Add Functional Tests

jordisala1991:
## Subject

I am targeting this branch, because this is BC.

Closes #669 

## Changelog

```markdown
### Fixed
- Fix ckeditor browser and upload actions.
```

PR #664, opened 2022-04-17
Fix addExtension all on formatters

VincentLanglet:
Solving https://github.com/sonata-project/SonataFormatterBundle/pull/645#issuecomment-1015256814 or https://github.com/sonata-project/SonataFormatterBundle/pull/659#issuecomment-1071728388

This is pretty weird ; seems like nobody is using those feature...

### Changelog

```markdown
## Fixed
- Adding extensions on formatters with config
```

PR #683, opened 2022-06-14
Improve Form Types

jordisala1991:
## Subject

I am targeting this branch, because this is in preparation for 5.0.

## Changelog

```markdown
### Added
- Add validations to form type options.

### Removed
- Remove `event_dispatcher` on FormatterType, since we provide it internally.
```

### DIFF
--- a/src/Form/Type/FormatterType.php
+++ b/src/Form/Type/FormatterType.php
@@ -16,7 +16,6 @@ namespace Sonata\FormatterBundle\Form\Type;
 use FOS\CKEditorBundle\Config\CKEditorConfigurationInterface;
 use Sonata\FormatterBundle\Form\EventListener\FormatterListener;
 use Sonata\FormatterBundle\Formatter\PoolInterface;
-use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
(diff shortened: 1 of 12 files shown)


tests/App/Admin/TextEntityAdmin.php
```diff
@@ -0,0 +1,38 @@
...
+final class TextEntityAdmin extends AbstractAdmin
+{
+    protected function configureFormFields(FormMapper $form): void
+    {
+        $form
+            ->add('simpleText', SimpleFormatterType::class, [
+                'format' => 'rawhtml',
+            ])
+            ->add('text', FormatterType::class, [
+                'listener' => false,
```
jordisala1991: If I use listener => true then I have to pass a event Dispatcher. The docs says this: https://docs.sonata-project.org/projects/SonataFormatterBundle/en/4.x/reference/formatter_widget/#how-to-use-it

Somehow the `$form` is suposed to have a `getEventDispatcher()`, but I don't see it.

@VincentLanglet do you know how to get an EventDispatcher here? Or do I have to inject one on the construct?

src/Form/Type/SimpleFormatterType.php
```diff
@@ -33,7 +33,10 @@ public function __construct(
     public function buildView(FormView $view, FormInterface $form, array $options): void
     {
         $defaultConfig = $this->ckEditorConfiguration->getDefaultConfig();
-        \assert(null !== $defaultConfig);
```
jordisala1991, line 36: This is not true, unless you configure on FOSCKEditorBundle a default configuration, this returns null: 

https://symfony.com/bundles/FOSCKEditorBundle/current/usage/config.html#define-a-configuration

```
The first configuration defined will be used as default configuration if you don't explicitly configure it.
```

src/Form/Type/FormatterType.php
```diff
@@ -189,7 +191,6 @@ public function configureOptions(OptionsResolver $resolver): void
             'ckeditor_plugins' => [],
             'ckeditor_templates' => [],
             'format_field_options' => $formatFieldOptions,
-            'source_field' => null,
```
jordisala1991, line 192: null is a value, so, the required fields do not explain that you need to set a value.

There is a lot to be improved on this form types.

jordisala1991:
https://github.com/sonata-project/dev-kit/pull/2076

jordisala1991:
Test are there, but changes on SRC could be moved to another PR. It could be worth to take a look at everything on the form folder, so we provide required fields and types for each form type.

tests/App/Resources/config/config.yaml
```diff
@@ -49,7 +61,12 @@ sonata_block:
...
 
 sonata_formatter:
-    default_formatter: rawhtml
+    default_formatter: text
     formatters:
+        text:
+            service: sonata.formatter.text.text
         rawhtml:
             service: sonata.formatter.text.raw
+            extensions:
```
jordisala1991: This could be added later, when we know how to test it.

tests/App/Admin/TextEntityAdmin.php
VincentLanglet: It's not `$form->getEventDispatcher()` but `$formBuilder->getEventDispatcher()`.
https://github.com/symfony/symfony/blob/6.0/src/Symfony/Component/Form/FormConfigBuilder.php#L182

So you can do `$form->getFormBuilder()->getEventDispatcher()`

jordisala1991: Fixed.

src/DependencyInjection/SonataFormatterExtension.php
```diff
@@ -152,16 +149,10 @@ private function createEnvironment(
...
-            && is_a($formatterDefinition->getClass(), ExtendableFormatter::class, true)
-        ) {
-            $formatterDefinition->addMethodCall('addExtension', [new Reference(sprintf('sonata.formatter.twig.sandbox.%s', $code))]);
+        $env->addMethodCall('addExtension', array(new Reference(sprintf('sonata.formatter.twig.sandbox.%s', $code))));
 
-            foreach ($extensions as $extension) {
-                $formatterDefinition->addMethodCall('addExtension', [new Reference($extension)]);
-            }
+        foreach ($extensions as $extension) {
+            $env->addMethodCall('addExtension', array(new Reference($extension)));
```
jordisala1991: This was changed like 2-3 times, from a working code to a non working code...

https://github.com/sonata-project/SonataFormatterBundle/pull/664
https://github.com/sonata-project/SonataFormatterBundle/pull/417

cc @VincentLanglet

src/Twig/Node/GistNode.php
```diff
@@ -31,9 +31,10 @@ public function __construct(
     public function compile(Compiler $compiler): void
     {
         $compiler
+            ->addDebugInfo($this)
             ->write(sprintf(
                 <<<EOT
-echo '<div class="sonata-gist"><script src="https://gist.github.com/%s.js?file=%s"></script></div>';\n,
+echo '<div class="sonata-gist"><script src="https://gist.github.com/%s.js?file=%s"></script></div>';\n
```
jordisala1991: This was changed from a working code to a non working one here:

https://github.com/sonata-project/SonataFormatterBundle/pull/305

Since 2017 this gist custom tag does not work.

src/DependencyInjection/SonataFormatterExtension.php
jordisala1991: Since 2019, adding extensions to formatters does not work.

jordisala1991:
IMO we should split changes on src on several PRs, and ensure with the test here that nothing breaks related to this breaks again. wdyt @VincentLanglet ?

src/DependencyInjection/SonataFormatterExtension.php
VincentLanglet: Yes this was reported. I thought I fixed this :s

VincentLanglet:
> IMO we should split changes on src on several PRs, and ensure with the test here that nothing breaks related to this breaks again. wdyt @VincentLanglet ?

Not sure to understand. I thought
- You added tests
- Tests wasn't working
- You fixed src code to get the tests working
So to me, this whole PR is good to be merged.

And it's better to merge the src changes with the tests rather than without.
Or do you mean you want to introduce some of this bugfix in 4.x ?

When I see that things like adding an extensions is not working since 2019 and was never reported I wonder if this worth the pain. But sure it would be better.

jordisala1991:
> > IMO we should split changes on src on several PRs, and ensure with the test here that nothing breaks related to this breaks again. wdyt @VincentLanglet ?
> 
> Not sure to understand. I thought
> 
> * You added tests
> * Tests wasn't working
> * You fixed src code to get the tests working
>   So to me, this whole PR is good to be merged.
> 
> And it's better to merge the src changes with the tests rather than without. Or do you mean you want to introduce some of this bugfix in 4.x ?
> 
> When I see that things like adding an extensions is not working since 2019 and was never reported I wonder if this worth the pain. But sure it would be better.

Test could be split so in a first iteration I add some tests and everything still works, and then I add the broken piece and I need to fix src.

Also there are other changes that are not "needed" to make tests pass.

Let me try, because I think we could split this somehow.

VincentLanglet:
https://github.com/sonata-project/SonataFormatterBundle/pull/679 is not enough for the browser action.
https://github.com/sonata-project/SonataFormatterBundle/blob/5.x/src/Resources/views/Ckeditor/browser.html.twig#L97 is not working since `sortparameters` doesn't exist in the modelManager (but the datagrid now).
Cf https://github.com/sonata-project/SonataFormatterBundle/pull/680#issuecomment-1157785630

jordisala1991:
After this one we could release an alpha for 5.0.0

VincentLanglet (approved):

PR #686, opened 2022-06-15
Stop using parameters on the container configuration

jordisala1991:
## Subject

I am targeting this branch, because this is BC break.

## Changelog

```markdown
### Removed
- All container parameters that were used to customize classes.
```


composer.json
```diff
@@ -42,7 +42,7 @@
         "twig/twig": "^2.12.1 || ^3.0"
     },
     "require-dev": {
-        "dama/doctrine-test-bundle": "^7.1",
+        "dama/doctrine-test-bundle": "^6.7",
```
VincentLanglet, line 45: Curious why ?

VincentLanglet (approved):

composer.json
jordisala1991: Class 'Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper' not found

on lowest deps, something on the lower side of deps is not correctly defined. I might need to take a deeper look.

jordisala1991: But for now, all other sonata bundles uses 6.7 so we can keep with it a little while.

PR #690, opened 2022-06-16
Add twig filter tests

jordisala1991:
## Subject

I am targeting this branch, because this is pedantic.

### DIFF
--- a/tests/App/Resources/views/text_formatter_test.html.twig
+++ b/tests/App/Resources/views/text_formatter_test.html.twig
@@ -1,0 +1,20 @@
+{#
+
+This file is part of the Sonata package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+
+#}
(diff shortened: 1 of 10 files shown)


tests/App/Resources/views/text_formatter_test.html.twig
```diff
@@ -0,0 +1,20 @@
...
+This file is part of the Sonata package.
+
+(c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+
+For the full copyright and license information, please view the LICENSE
+file that was distributed with this source code.
+
+#}
+
+{%- apply spaceless|format_text('rawhtml')|raw %}
```
jordisala1991, line 12: Not sure if having to use raw is desired here, IMO we should specify that this filter might produce html so it gets raw already. wdyt @VincentLanglet (for another Pr obv)

jordisala1991: Will be solved with next PR: #691 , but first we need this one merged.

VincentLanglet (approved):

PR #693, opened 2022-06-17
Add more functional tests for browser

jordisala1991:
## Subject

I am targeting this branch, because this is broken on NEXT MAJOR.

Closes #680.


jordisala1991:
Here I might be able to reproduce the error.

jordisala1991:
@t-works @VincentLanglet in this PR we should address all errors on the browser template.

Let's leave the possible optimization if they do not throw an error or show incorrect data to the user, because I would like to try instead of using the empty layout, try and use the base_list layout. Here we are copying a lot of code from SonataAdminBundle and it is not ideal.

2 steps process:
1. Let's try to make this work
2. Let's see if we can use SonataAdminBundle base_list with some tweaks.
3. (optional) if we can't, let's see what we can optimize here to avoid all the html duplicated.

src/Resources/views/Ckeditor/browser.html.twig
```diff
@@ -92,17 +91,19 @@ file that was distributed with this source code.
...
+                                            {% set sort_by              = current ? admin.datagrid.values[constant('Sonata\\AdminBundle\\Datagrid\\DatagridInterface::SORT_ORDER')] : field_description.option(constant('Sonata\\AdminBundle\\Datagrid\\DatagridInterface::SORT_ORDER'), sort_parameters.filter[constant('Sonata\\AdminBundle\\Datagrid\\DatagridInterface::SORT_ORDER')]) %}
                                         {% endif %}
 
                                         {% apply spaceless %}
                                             <th class="sonata-ba-list-field-header-{{ field_description.type}} {% if sortable %} sonata-ba-list-field-header-order-{{ sort_by|lower }} {{ sort_active_class }}{% endif %}">
-                                                {% if sortable %}<a href="{{ admin.generateUrl('ckeditor_browser', sort_parameters) }}">{% endif %}
+                                                {% if sortable %}<a href="{{ admin.generateUrl(action, sort_parameters) }}">{% endif %}
                                                     {{ field_description.label|trans }}
```
VincentLanglet:
```
{% if field_description.label is not same as(false) %}
                                                        {% if field_description.translationDomain is same as(false) %}
                                                            {{ field_description.label }}
                                                        {% else %}
                                                            {{ field_description.label|trans({}, field_description.translationDomain) }}
                                                        {% endif %}
                                                    {% endif %}
```

src/Resources/views/Ckeditor/browser.html.twig
```diff
@@ -92,17 +91,19 @@ file that was distributed with this source code.
                                         {# Disable batch and actions #}
```
VincentLanglet: Use above
```
constant('Sonata\\AdminBundle\\Datagrid\\ListMapper::NAME_BATCH')
constant('Sonata\\AdminBundle\\Datagrid\\ListMapper::NAME_ACTIONS')
``

VincentLanglet (commented):
I cannot access to the admin.datagrid.result|length > 0 code to comment.

We can use the same tips than https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Resources/views/CRUD/base_list.html.twig#L33-L34 for performance improvements.

VincentLanglet:
> Let's leave the possible optimization if they do not throw an error or show incorrect data to the user, because I would like to try instead of using the empty layout, try and use the base_list layout. Here we are copying a lot of code from SonataAdminBundle and it is not ideal.
>
> 2 steps process:
> 
> Let's try to make this work
> Let's see if we can use SonataAdminBundle base_list with some tweaks.
> (optional) if we can't, let's see what we can optimize here to avoid all the html duplicated.

I'm not sure it's possible but removing the duplicated code would be awesome

src/Resources/views/Ckeditor/browser.html.twig
jordisala1991: Applied

jordisala1991:
> > Let's leave the possible optimization if they do not throw an error or show incorrect data to the user, because I would like to try instead of using the empty layout, try and use the base_list layout. Here we are copying a lot of code from SonataAdminBundle and it is not ideal.
> > 2 steps process:
> > Let's try to make this work
> > Let's see if we can use SonataAdminBundle base_list with some tweaks.
> > (optional) if we can't, let's see what we can optimize here to avoid all the html duplicated.
> 
> I'm not sure it's possible but removing the duplicated code would be awesome

There you have it, have a check. I have tested it with one of my projects and it works so far. (I am not using SonataClassificationBundle, but should work since I checked this controller is mostly the same as the SonataMediaBundle list controller

VincentLanglet (approved):

PR #694, opened 2022-06-17
Removed useless class

jordisala1991:
## Subject

I am targeting this branch, because this is for NEXT MAJOR.

## Changelog

```markdown
### Removed
- Remove `Sonata\FormatterBundle\Extension\BaseProxyExtension`.
```


src/Extension/BaseProxyExtension.php
```diff
@@ -1,95 +0,0 @@
-<?php
```
jordisala1991, line 1: This class was used on SonataMediaBundle a long time ago:

https://github.com/sonata-project/SonataMediaBundle/commit/4881f38183c3ede83b083f29044e85e94293bfbf

But the usage was removed when we tried to remove the double dependency (Media depends on formatter and formatter depends on media):

https://github.com/sonata-project/SonataMediaBundle/pull/1478

With that change, the media extension that can render images inside a formatter field became completely broken (nice!).

TODAY:

1. This class is no longer needed since now we use Runtimes from Twig.
2. We should add the Media extension here instead of having it on MediaBundle (next PR)
3. We should remove this file: https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Twig/Extension/FormatterMediaExtension.php which is completely broken.

jordisala1991: I have checked across all sonata code and it is not used anymore.

VincentLanglet (approved):

PR #695, opened 2022-06-17
Update deps

jordisala1991:
## Subject

I am targeting this branch, because this is for next major.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,8 +42,8 @@
         "twig/twig": "^2.12.1 || ^3.0"
     },
     "require-dev": {
-        "dama/doctrine-test-bundle": "^6.7",
-        "doctrine/doctrine-bundle": "^2.3.2",
+        "dama/doctrine-test-bundle": "^7.1",
+        "doctrine/doctrine-bundle": "^2.4.3",
         "doctrine/orm": "^2.9",
         "friendsofphp/php-cs-fixer": "^3.4",
         "knplabs/knp-menu-bundle": "^3.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -42,8 +42,8 @@
         "twig/twig": "^2.12.1 || ^3.0"
     },
     "require-dev": {
-        "dama/doctrine-test-bundle": "^6.7",
-        "doctrine/doctrine-bundle": "^2.3.2",
+        "dama/doctrine-test-bundle": "^7.1",
```
jordisala1991: If we use ^7.1 this forces to use dbal 3. wdyt @VincentLanglet ?

We have some 4.x versións of Sonata still allowing dbal 2. should we drop it already or we wait?

VincentLanglet: Maybe we should wait.

This major will allow people to use symfony 5. People might need to first bump some dependencies before bumping dbal.

jordisala1991:
```suggestion
        "dama/doctrine-test-bundle": "^6.7",
```

composer.json
```diff
@@ -42,8 +42,8 @@
...
+        "doctrine/doctrine-bundle": "^2.4.3",
```
jordisala1991:
```suggestion
        "doctrine/doctrine-bundle": "^2.3.2",
```

composer.json
jordisala1991: Fine by me, then lets only update the psalm extension

VincentLanglet (approved):

PR #696, opened 2022-06-17
Add media extension for formatter

jordisala1991:
## Subject

I am targeting this branch, because this is in preparation for 5.0.

## Changelog

```markdown
### Added
- Added `MediaExtension` to render Media inside Formatters.
- Added ability to use Twig runtimes on your Formatter Extensions via `getAllowedRuntimes`.
```

### DIFF
--- a/src/DependencyInjection/SonataFormatterExtension.php
+++ b/src/DependencyInjection/SonataFormatterExtension.php
@@ -15,8 +15,10 @@ namespace Sonata\FormatterBundle\DependencyInjection;
 
 use Sonata\FormatterBundle\Twig\Loader\LoaderSelector;
 use Sonata\FormatterBundle\Twig\SecurityPolicyContainerAware;
+use Sonata\MediaBundle\Twig\MediaRuntime;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
@@ -26,6 +28,7 @@ use Twig\Environment;
 use Twig\Extension\SandboxExtension;
 use Twig\Lexer;
 use Twig\Loader\ArrayLoader;
+use Twig\RuntimeLoader\ContainerRuntimeLoader;
 
 /**
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
@@ -59,7 +62,7 @@ final class SonataFormatterExtension extends Extension
         }
 
         if (isset($bundles['SonataMediaBundle'])) {
-            $loader->load('ckeditor.php');
+            $loader->load('media.php');
         }
 
         if (!\array_key_exists($config['default_formatter'], $config['formatters'])) {
@@ -119,6 +122,9 @@ final class SonataFormatterExtension extends Extension
         string $code,
         array $extensions
     ): string {
+        $bundles = $container->getParameter('kernel.bundles');
+        \assert(\is_array($bundles));
+
         $loader = new Definition(ArrayLoader::class);
 
         $loader->setPublic(false);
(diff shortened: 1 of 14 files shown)


src/DependencyInjection/SonataFormatterExtension.php
```diff
@@ -151,6 +157,18 @@ private function createEnvironment(
 
         $env->addMethodCall('addExtension', [new Reference(sprintf('sonata.formatter.twig.sandbox.%s', $code))]);
 
+        if (isset($bundles['SonataMediaBundle'])) {
+            $runtimeLoader = new Definition(ContainerRuntimeLoader::class, [
+                ServiceLocatorTagPass::register($container, [
+                    MediaRuntime::class => new Reference('sonata.media.twig.runtime'),
```
jordisala1991: I do not like to do it this way, because user could add its own extensions needing own custom runtime.

jordisala1991:
WDYT @VincentLanglet ? I had to add some stuff to make it work with Runtimes because otherwise don't get loaded by those custom Twig Environments.

src/Extension/ExtensionInterface.php
```diff
@@ -39,4 +41,9 @@ public function getAllowedProperties(): array;
      * @return array<class-string, array<string>>
      */
     public function getAllowedMethods(): array;
+
+    /**
+     * @return array<string, class-string<RuntimeExtensionInterface>>
```
VincentLanglet: It might worth to explain the shape of the array. Keys need to be the service ids, it's not obvious.

VincentLanglet (dismissed):
Can you explain me what problem do we solve here ?
twig extensions/runtimes from MediaBundle were not loaded ?

I dunno if there is docs about extensions, but this might be interesting to add some. Especially because we're adding methods.

But overall the implementation seems ok to me

jordisala1991:
There are 2 things in this PR:

1. Trying to solve the MediaExtension issue: It is implemented on MediaBundle but completely broken. It should be implemented here.

2. With Runtimes of twig we can separate extensions and their logics. The problem is that we cant use them on formatters if we dont configure the Environment. Currently without this Pr there is no way someone can add a custom extension for formatters is those use Runtimes (like the media extension does).

If we agree on implementation I can add missing docs and  polish it a little bit.

src/DependencyInjection/SonataFormatterExtension.php
```diff
@@ -119,6 +122,9 @@ private function createEnvironment(
         string $code,
         array $extensions
     ): string {
+        $bundles = $container->getParameter('kernel.bundles');
+        \assert(\is_array($bundles));
+
```
jordisala1991:
```suggestion
```

src/Extension/ExtensionInterface.php
jordisala1991: We might only provide the Id and I can retrieve that Id from the container to get the class name, that would be better?

VincentLanglet: All the methods are returning string[] or string[][]. So It might be better to have a similar signature indeed.

VincentLanglet:
> If we agree on implementation I can add missing docs and polish it a little bit.

Except our little discussion, I'm fully ok with it.

src/DependencyInjection/SonataFormatterExtension.php
jordisala1991: Fixed.

src/Extension/ExtensionInterface.php
jordisala1991: Explained

tests/custom_bootstrap.php
```diff
@@ -35,3 +35,9 @@
     'command' => 'doctrine:schema:create',
 ]);
 $application->run($input, new NullOutput());
+
+$input = new ArrayInput([
```
jordisala1991, line 39: we should add this everywhere to improve coverage , otherwise some classes get executed only on the previous commands.

VincentLanglet (approved):

tests/custom_bootstrap.php
VincentLanglet: Do you plan to do the PR or should I ?

VincentLanglet:
Is it ready to merge ?

VincentLanglet:
If I understand correctly, we should provide a PR to deprecate https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Twig/Extension/FormatterMediaExtension.php isn't it ?

jordisala1991:
I would say we should remove it directly. That class does nothing and it is not registered on the container.

VincentLanglet:
> I would say we should remove it directly. That class does nothing and it is not registered on the container.

Removing something is theoretically a BC-break, but this seems ok to me.

tests/custom_bootstrap.php
jordisala1991: If you can do them it would be great

PR #720, opened 2022-09-13
fix: change list_outer_rows_browser.html.twig template name.

m-ar-c:
## Subject

`outer_list_rows_browser` does not match any current template name in Sonata. With `outer_list_rows_list` instead, the
[list_outer_rows_browser.html.twig](https://github.com/sonata-project/SonataFormatterBundle/blob/5.x/src/Resources/views/Ckeditor/list_outer_rows_browser.html.twig) template is used and the javascript in [browser.html.twig](https://github.com/sonata-project/SonataFormatterBundle/blob/5.x/src/Resources/views/Ckeditor/browser.html.twig) is working again (we can select images by clicking on them).

I am targeting this branch, because it is a BC fix.

## Changelog

```markdown
### Fixed
- Fix wrong template name preventing users to select images.
```

### DIFF
--- a/src/Admin/CkeditorAdminExtension.php
+++ b/src/Admin/CkeditorAdminExtension.php
@@ -28,7 +28,7 @@ final class CkeditorAdminExtension extends AbstractAdminExtension
 {
     public function configure(AdminInterface $admin): void
     {
-        $admin->setTemplate('outer_list_rows_browser', '@SonataFormatter/Ckeditor/list_outer_rows_browser.html.twig');
+        $admin->setTemplate('outer_list_rows_list', '@SonataFormatter/Ckeditor/list_outer_rows_browser.html.twig');
     }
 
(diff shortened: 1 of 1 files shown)


jordisala1991:
I don't think this change is correct, test are failing.

m-ar-c:
Right, this is not correct. I had the following
``` php
    public function configureListFields(ListMapper $list): void
    {
            $list->getAdmin()->setListMode('list');
    }
```
in an other Admin Extension also targeting `sonata.media.admin.media`. This was responsible for the bug.

I changed it to :

``` php
    public function configureListFields(ListMapper $list): void
    {
            if ('browser' !== $list->getAdmin()->getListMode())
                $list->getAdmin()->setListMode('list');
    }
```

Now everything works as it should. Sorry for the noise.

